Issue #16, opened 2020-06-24
Design System: Enhance our CodeBlock component.


jimmylee:
# Purpose

This issue is intended as an introduction to web development/React task. There exists a `CodeBlock` component used in the design system that I would like beautified.

The file lives here: https://github.com/filecoin-project/slate/blob/master/components/system/components/CodeBlock.js

# Deliverable

+ Start with tweaks and adjustments, and then maybe more incremental improvements such as line number, etc.
+ Getting comfortable with pull requests, and other software development process over Github.

# Dependencies

N/A

jimmylee:
Making a note here that I want syntax highlighting now. Only `TypeScript` and `JavaScript` needs to be supported.

jimmylee:
Adding @gndclouds to work on this as an intro task.

PR #22, opened 2020-06-29
design system: update code block component

tarafanlin:
https://github.com/filecoin-project/slate/issues/16
this is an example of a pull request


jimmylee:
This was a great exercise to get a pull request started on the repo! @tarafanlin you learned so quickly, nice work. Merging this now and then we will make more changes later to satisfy the requirements in the issue.

components/system/components/CodeBlock.js
```diff
@@ -1,13 +1,15 @@
-import * as React from 'react';
-import * as Constants from '~/common/constants';
+import * as React from "react";
+import * as Constants from "~/common/constants";
 
-import { css } from '@emotion/react';
+import { css } from "@emotion/react";
```
jimmylee, line 4: LGTM! We will want to make sure your prettier config is updated at some point so our formatting stays the same.

PR #29, opened 2020-07-01
New notification component and active link highlighting

martinalong:
- Created Notification component with 4 modes, option to add description and label, and option to include an onClose function
- Changed styling of input with validation
- Added active-link highlighting so it's easier to see which page you're on
- Plus misc spelling / formatting edits

image

image

image

### DIFF
--- a/components/system/SystemPage.js
+++ b/components/system/SystemPage.js
@@ -67,10 +67,15 @@ const STYLES_LINK = css`
   :hover {
     color: ${Constants.system.brand};
   }
+`;
 
-  :visited {
-    color: ${Constants.system.black};
-  }
+const STYLES_LINK_ACTIVE = css`
+  font-family: 'inter-semi-bold';
+  color: ${Constants.system.brand};
+  text-decoration: none;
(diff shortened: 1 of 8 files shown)


components/system/SystemPage.js
```diff
@@ -90,9 +95,10 @@ const STYLES_LABEL = css`
 `;
 
 const SidebarLink = (props) => {
+  console.log(props.url)
```
jimmylee, line 98: nit: Not a big deal, but make sure you delete these console statements later.

components/system/SystemPage.js
```diff
@@ -90,9 +95,10 @@ const STYLES_LABEL = css`
...
   return (
     <React.Fragment>
-      <a css={STYLES_LINK} href={props.href} target={props.target}>
+      <a css={props.url.includes(props.href) ? STYLES_LINK_ACTIVE : STYLES_LINK} href={props.href} target={props.target}>
```
jimmylee, line 101: This is really clever! I had not thought about this.

Issue #37, opened 2020-07-01
Adding arguments to `npm run dev` so we don't delete the local settings and data each time.


jimmylee:
# Purpose

Before when I was working on `slate` alone, I would delete all of the contents of `.data` each time so I got the new user experience every single time I opened it.

However, this can be suboptimal for a number of reasons:

+ You may not want to reset your state
+ You may want to keep your files
+ You may want to test an edge case that only occurs a certain state
+ You may want to borrow someone else's `/public/static` and `/.data` so you can easily reproduce any bug that someone is experiencing (the superpower of our current configuration).

# Deliverable

+ Make `npm run dev` accept arguments, where `npm run dev --reset-data` is accepted.

# Notes

+ This is a good lead: https://www.twilio.com/blog/npm-scripts then scroll to **Argument passing and parsing**
+ https://stackoverflow.com/questions/51388921/pass-command-line-args-to-npm-scripts-in-package-json
+ You can see how I do it for `lotus` here: https://github.com/filecoin-project/lotus-docs (another good example codebase)

jimmylee:
Thank you @akuokojnr 🙏

PR #49, opened 2020-07-07
feat: add argument to enable the ability to clear local files and settings

akuokojnr:
This PR fixes [Issue #37](https://github.com/filecoin-project/slate/issues/37)

Currently, when `npm run dev` is ran, all contents of `.data` are deleted. This presents a new user experience every single time the development command is ran. 

This PR modifies `server.js` to enable the user to add the boolean flag `--reset-data` to `npm run dev` as option to either clear or keep all local files and settings. 

The default behavior now is to present the previous state to the user. If the user wishes to clear the previous data, the user can run `npm run dev --reset-data`

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ npm run dev
 
 - Visit `localhost:1337` in the browser.
 - If you see the design system page instead, that means a token was not properly generated in the `.data` folder. Run `npm run dev` again.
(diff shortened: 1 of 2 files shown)


jimmylee:
I love this @akuokojnr, thank you so much 🙏

PR #60, opened 2020-07-10
added FFS PowerGate token to fixtures.js

johannes-jp:
1 of 6 tokens for PowerGate calls in server.js

### DIFF
--- a/common/fixtures.js
+++ b/common/fixtures.js
@@ -294,3 +294,36 @@ export const SELECT_COUNTRY_OPTIONS = [
   { value: "Zimbabwe", name: "Zimbabwe" },
   { value: "Åland Islands", name: "Åland Islands" },
 ];
+
+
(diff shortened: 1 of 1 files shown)


jimmylee:
This is really great, one thing that I would say that I didn't get a chance to say earlier:

```json
{
   id: 123123132131321,
   token: 123211312321312313
}
```

Keeping the object syntax is preferred, because then its copy & paste friendly. For example If I just wanted to copy and paste it and test it as a property for a component, I would do

```js
const testValue = {
   id: 123123132131321,
   token: 123211312321312313
}
```

And you keep that plug and play capability.

I'm going to merge this PR. I'd like you to make a new one and continue the work from there! Thanks :D

johannes-jp:
Tracking, I've adjusted the formatting which you'll see reflected in the next PR. Thanks for the feedback!

PR #65, opened 2020-07-14
GlobalNotification + CodeText refactor

martinalong:
- Refactored notifications to be global and updated documentation page
- Refactored code text in tables to be a specific TableComponent for easier usage

image
image


components/system/components/GlobalNotification.js
```diff
@@ -0,0 +1,108 @@
...
+import * as Constants from "~/common/constants";
+import * as SVG from "~/components/system/svg";
+import * as Strings from "~/common/strings";
+
+import { css } from "@emotion/react";
+
+import { DescriptionGroup } from "~/components/system/components/fragments/DescriptionGroup";
+
+const STYLES_NOTIFICATION_LIST = css`
+  position: fixed;
```
jimmylee, line 11: I don't believe `display: flex;` and `flex-direction: column` are needed at all here. In fact it is redundant because block elements are naturally stacked on top of each other in a flow based layout system (How the DOM naturally behaves with `inline`, `inline-block`, and `block` elements. So basically all you need is position fixed.

There is also nothing wrong with having some default `right: 0; bottom: 0;` on this element since `inline-styles` always take priority over class styles.

pages/system/notifications.js
```diff
@@ -40,182 +45,373 @@ export default class SystemPageNotifications extends React.Component {
...
-          description="Here is the description"
-          status="SUCCESS"
-        />
+        <System.P>
+          Use <System.CodeText>style</System.CodeText> to specify placement of
+          the fixed positioning notification list
+        </System.P>
+        <br />
+        <System.CodeBlock>
+          {`function App() {
```
jimmylee, line 70: Make this example a `React` class with a `render()` method.

pages/system/notifications.js
```diff
@@ -40,182 +45,373 @@ export default class SystemPageNotifications extends React.Component {
...
+              dark: true,
+            })
+          }
+        >
+          Click for dark style notification
+        </System.ButtonSecondaryFull>
+        <br />
+        <System.ButtonPrimaryFull
+          onClick={() => {
+            for (let i = 0; i <= this.state.count; i++) {
```
jimmylee, line 111: I would prefer instead of a call to delete each notification you just had a way to assign a new empty array. Did you do it this way because you wanted to see each transition out?

components/system/components/GlobalNotification.js
```diff
@@ -0,0 +1,108 @@
+import * as React from "react";
+import * as Constants from "~/common/constants";
+import * as SVG from "~/components/system/svg";
+import * as Strings from "~/common/strings";
+
+import { css } from "@emotion/react";
+
```
jimmylee, line 7: Extra new line ✂️

pages/system/notifications.js
martinalong: I assumed that people might want notifications that appeared at different times to disappear at different times, but I can definitely just assign an empty array if you think that's better. That would probably be easier too

jimmylee: I think that is a lot better and simplifies the count.

Not sure why `this.state.count` is necessary, if its an Object you can do `Object.keys(object).length` or if its an array you can do `array.length`

PR #68, opened 2020-07-14
Global modals

martinalong:
- Global modals and documentation added
- Edit to notification handler function names ( `_createNotification` -> `_handleCreate`) as suggested

image

### DIFF
--- a/components/system/components/GlobalModal.js
+++ b/components/system/components/GlobalModal.js
@@ -7,8 +7,8 @@ import { css } from "@emotion/react";
 
 const STYLES_BACKGROUND = css`
   position: fixed;
-  left: 0;
-  top: 0;
+  right: 0;
+  bottom: 0;
   width: 100%;
   height: 100%;
   display: flex;
--- a/components/system/components/GlobalModal.js
+++ b/components/system/components/GlobalModal.js
@@ -1,0 +1,77 @@
+import * as React from "react";
+import * as Constants from "~/common/constants";
+import * as SVG from "~/components/system/svg";
+import * as Strings from "~/common/strings";
+
+import { css } from "@emotion/react";
+
+const STYLES_BACKGROUND = css`
+  position: fixed;
+  left: 0;
+  top: 0;
+  width: 100vw;
+  height: 100vh;
+  background-color: rgba(45, 41, 38, 0.6);
+  z-index: ${Constants.zindex.modal};
+`;
+
+const STYLES_MODAL = css`
+  width: 50vw;
+  height: 60vh;
+  position: absolute;
+  top: 50%;
+  left: 50%;
+  transform: translate(-50%, -50%);
+  border-radius: 4px;
+  background-color: ${Constants.system.white};
+`;
+
+const STYLES_X = css`
+  height: 24px;
+  position: absolute;
+  top: 8px;
+  right: 8px;
+  cursor: pointer;
+`;
+
+export class GlobalModal extends React.Component {
+  state = {
+    modal: null,
+  };
+
+  componentDidMount = () => {
+    window.addEventListener("create-modal", this._handleCreate);
+    window.addEventListener("delete-modal", this._handleDelete);
+  };
+
+  componentWillUnmount = () => {
+    window.removeEventListener("create-modal", this._handleCreate);
+    window.removeEventListener("delete-modal", this._handleDelete);
+  };
+
+  _handleCreate = (e) => {
+    this.setState({ modal: e.detail.modal });
+  };
+
+  _handleDelete = (e) => {
+    this.setState({ modal: null });
+  };
+
+  render() {
+    if (this.state.modal) {
+      return (
+        <div
+          css={STYLES_BACKGROUND}
+          style={this.props.backgroundStyle}
+          onClick={this._handleDelete}
+        >
+          <div css={STYLES_MODAL} style={this.props.style}>
+            {this.state.modal}
+            <SVG.X css={STYLES_X} onClick={this._handleDelete} />
+          </div>
+        </div>
+      );
+    }
+    return null;
+  }
+}
--- a/components/system/components/GlobalModal.js
+++ b/components/system/components/GlobalModal.js
@@ -7,8 +7,10 @@ import { css } from "@emotion/react";
 
 const STYLES_BACKGROUND = css`
   position: fixed;
+  left: 0;
   right: 0;
   bottom: 0;
+  top: 0;
   width: 100%;
   height: 100%;
   display: flex;
@@ -23,8 +25,6 @@ const STYLES_MODAL = css`
   padding: 8px;
   max-width: 568px;
   width: 100%;
-  max-height: 568px;
-  height: 100%;
   border-radius: 4px;
   background-color: ${Constants.system.white};
 `;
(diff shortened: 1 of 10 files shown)


components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+import * as SVG from "~/components/system/svg";
+import * as Strings from "~/common/strings";
+
+import { css } from "@emotion/react";
+
+const STYLES_BACKGROUND = css`
+  position: fixed;
+  left: 0;
+  top: 0;
+  width: 100vw;
```
jimmylee, line 12: `width: 100%;` `right: 0;` `bottom: 0;`

components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+  height: 100vh;
```
jimmylee, line 13: ✂️  remove this `height: 100vh`, go with `display: flex; align-items: center; justify-content: center;` and all children will be centered, you won't need to have a modal thats absolute positioned as the child.

components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+  background-color: rgba(45, 41, 38, 0.6);
+  z-index: ${Constants.zindex.modal};
+`;
+
+const STYLES_MODAL = css`
+  width: 50vw;
+  height: 60vh;
+  position: absolute;
```
jimmylee, line 21: `max-width: 568px; width: 100%;` this will allow the modal to be fluid in smaller width contexts. (free mobile responsiveness)

jimmylee, line 21: ✂️ no need for position absolute, its in a fixed element container. (slower paint)

components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+  top: 50%;
```
jimmylee, line 22: cut all `top, left, transform` rules.

components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+  left: 50%;
+  transform: translate(-50%, -50%);
+  border-radius: 4px;
+  background-color: ${Constants.system.white};
+`;
+
+const STYLES_X = css`
```
jimmylee, line 29: We should use a better name for this, not a big deal this pass but this won't feel good when you come back to it in the future.

components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+  _handleCreate = (e) => {
+    this.setState({ modal: e.detail.modal });
+  };
+
+  _handleDelete = (e) => {
+    this.setState({ modal: null });
+  };
+
+  render() {
+    if (this.state.modal) {
```
jimmylee, line 61: Stylistically, sometimes its more attractive to re-order this to

```js

if (!negatedCondition) {
  return null;
}

return (
   <div>...</div>
)
```

In addition you can do

```js
if (!negatedState) return null;

return (
   <div>...</div>
);
```

components/system/components/GlobalModal.js
```diff
@@ -0,0 +1,77 @@
...
+      return (
+        <div
+          css={STYLES_BACKGROUND}
+          style={this.props.backgroundStyle}
+          onClick={this._handleDelete}
+        >
+          <div css={STYLES_MODAL} style={this.props.style}>
+            {this.state.modal}
+            <SVG.X css={STYLES_X} onClick={this._handleDelete} />
```
jimmylee, line 70: We may already have a Dismiss SVG. Can you double check for me? Thank you!

pages/system/modals.js
```diff
@@ -0,0 +1,238 @@
...
+          style the background and <System.CodeText>style</System.CodeText> to
+          style the modal itself.
+        </System.P>
+        <br />
+        <System.CodeBlock>
+          {`class App extends React.Component {
+  render() {
+    return(
+      <React.Fragment>
+        <GlobalModal style={{ padding: "10px" }}/>
```
jimmylee, line 71: Maybe we don't need `10px` here? Lets just remove this from the example and add that 10 (preferably something base 8 instead) to the default case.

pages/system/modals.js
```diff
@@ -0,0 +1,238 @@
...
+        <hr />
+        <br />
+        <System.ButtonSecondaryFull
+          onClick={() =>
+            this._handleCreate({
+              modal: (
+                <div
+                  style={{
+                    textAlign: "center",
+                    position: "absolute",
```
jimmylee, line 92: Hmm I find this absolute nesting to be odd, You already have the modal in a fixed container, there shouldn't be an absolute nesting.

martinalong: I was using it moreso to vertically+ horizontally center the text and button inside the modal. When I remove the `position: absolute` and leave it as 
image
things start to get a bit wonky. I did it this way because `margin: auto` didn't center it vertically. Let me know if you know a better way to center it!
image

components/system/components/GlobalModal.js
martinalong: I didn't find it in "~/common/svg" or "~/components/system/svg". I'll rename it from X to Close though

martinalong:
Do you have a preference for height of the modal? I just made it the same as how the width is set for now

image

martinalong:
Just curious why `right: 0` ` bottom: 0` is preferred to `left: 0` `top: 0` ?

jimmylee:
To answer your min-height comment. The height should be dynamic based on the children that are in the element. I'm going to provide you a codepen that answers this later.

You want to be specific that you use `right` `bottom` `left` and `top` 0 to guarantee it draws properly to all the edges. In this case you don't necessarily need to specify a width either (I do for good measure). (This is for width: 100%)

There is also a third thing we can do after this PR lands which is a utility component to handle outside clicks of content to hide the modal. (Will be fun for you to learn)

jimmylee:
I also posted in Slack, but I'm posting here too because I'm repetitive 😆 

Here is a Codepen that demonstrates some of the simplifications I'd like to the Overlay/Modal. 

https://codepen.io/meanjim/pen/dyGqJQw

* Not specifying a width for a fixed element gives us more flexibility on how we space the pixels against the top, left, right, and bottom edges. Also reduces the style complexity. **Try playing with the values to see what I mean!**
* As you can see, for the child we don't have to specify very many rules. However we should make sure that the child is position:relative so absolute elements that are children of this child will be positioned relatively to it.
* I think its worthwhile to add some of the redundant properties such as `box-sizing: border-box;` since when they import `slate-react-system` they won't have our global overrides.

Thanks @martinalong !!

martinalong:
Ahh ok that makes a lot of sense. Thanks for that example! And I wish I'd known about box-sizing: border-box sooner haha. I've encountered that problem a lot with overflowing elements but never knew there was a styling to fix it.

pages/system/modals.js
```diff
@@ -0,0 +1,237 @@
+import * as React from "react";
+import * as System from "~/components/system";
+import Group from "~/components/system/Group";
```
jimmylee, line 3: Nit: This isn't a big deal, but lets keep imports grouped by style for less inference pain when you first read a file.

PR #73, opened 2020-07-19
fixed margins and padding of table and drag icon

gndclouds:
Made some tiny tweaks to center margins and padding for tables and drag icon in list editor.

Drag Icon Margins
Group 1

Table Padding
Group 2


jimmylee:
@gndclouds this is super helpful, thank you !!

PR #75, opened 2020-07-20
Bugfix: strings.bytesToSize method returns inaccurate decimal places

motdde:


common/strings.js
```diff
@@ -49,7 +49,7 @@ export const bytesToSize = (bytes, decimals = 2) => {
 
   const i = Math.floor(Math.log(bytes) / Math.log(k));
 
-  return parseFloat((bytes / Math.pow(k, i)).toFixed(dm)) + " " + sizes[i];
+  return `${(bytes / Math.pow(k, i)).toFixed(dm)}  ${sizes[i]}`;
```
jimmylee, line 52: Thanks @motdde !!

motdde: Thanks @jimmylee

PR #82, opened 2020-07-20
Fixed table formatting bug

martinalong:
Expandable dropdown component now no longer creates an offset in the column titles for non-expandable tables

image


jimmylee:
LGTM :)

Issue #77, opened 2020-07-20
Design: Marketing pages.


jimmylee:
# Purpose

* @gndclouds and @tarafanlin have done some pretty cool stuff with our Typography guidelines. And things will keep getting better and better over time.
* We are lucky to have @harisbutt on the team. Our narrative for Slate should be even stronger.
* This issue describes how some of the vision, brand, and thinking manifests into real deliverables.

# Deliverable

* V0 of Landing, Community Guidelines, Terms, Privacy

# Dependencies

* https://github.com/filecoin-project/slate/issues/11
* https://github.com/filecoin-project/slate/issues/76
* https://github.com/filecoin-project/slate/issues/70
* https://github.com/filecoin-project/slate/issues/56

gndclouds:
Are there any existing sketch or Figma docs we should be starting from?

gndclouds:
Started a [Figma file](https://www.figma.com/file/U26YdNSJLWV8JK7h4YQl2I/Slate?node-id=0%3A1) to brainstorm a bit. Ping me if you have issues editing it.

There are currently four pages:
- 🎨 Marketing Site
- 📐 Design System Docs related to #87 
- 🧠 Brainstorming
- 🤞🏻Diagrams and Systems

gndclouds:
New [Figma File Link](https://www.figma.com/file/DYMWe6q2zidE40zt63VqFD/Landing-Pages?node-id=0%3A1)

tarafanlin:
### Collaboration
- [ ] top slate block
- [ ] transition animation

###  @gndclouds sections
- [ ] create slate section
- [ ] share section
- [ ] free, unlimited storage

###  @jordattebayo sections
- [ ] simple, intuitive section
- [ ] open source

### @tarafanlin sections
- [ ] private & secure 
- [ ] store files
- [ ] footer

tarafanlin:
> ### Collaboration
> * [ ]  top slate block
> * [ ]  transition animation
> * [ ]  chrome pluin section
> 
> ### @gndclouds sections
> * [x]  create slate
> * [x]  share
> * [x]  free, unlimited storage (Pausing this section for now.)
> 
> ### @jordattebayo sections
> * [ ]  simple, intuitive
> * [ ]  open source
> 
> ### @tarafanlin sections
> * [x]  private & secure
> * [x]  store files
> * [x]  footer

Finished my part for the first pass. 
I will start working on the top slate block animation if you don't need any assistance~

[...]

PR #85, opened 2020-07-21
@gndclouds/typography

gndclouds:
We have a typography page (!) which we can continue to advance as page of #77 but here is the first iteration.


jimmylee:
LGTM :)

WHAT A GREAT START!

Issue #50, opened 2020-07-07
 Design: Update the create Powergate token component to have an odometer effect.


jimmylee:
# Purpose

<img width="1293" alt="Screen Shot 2020-07-06 at 8 30 34 PM" src="https://user-images.githubusercontent.com/310223/86702186-935d8c00-bfc7-11ea-8adb-71f56da6fb43.png">

* There are developers that will use this example to plug and play Textile's powergate. Lets make it more fun!
* An odometer effect from `XXXXXXXX-XXXX-XXXX-XXXX-XXXXXXXXXXXX` to `82324ce3-3fb4-4eea-ac75-f7797d96a403` would be really cool to see.
* Perfect for the kind of developer that loves to see a little fun :)

# Deliverable

* An updated component with an odometer effect.

jimmylee:
@akuokojnr this turned out so good.

PR #88, opened 2020-07-21
feat: add odometer effect to create token component

akuokojnr:
This PR fixes Issue #50. It adds an odometer effect to the create Powergate token component.


jimmylee:
LGTM. This is great, we'll improve it over time too.

PR #110, opened 2020-07-23
added consistency and a new color to link styles

tmytrn:
Here's my proposed changes to add better link styles. 

I read the @emotion docs about [style composition](https://emotion.sh/docs/composition), but had an error`Attempted import error: '@emotion/css' does not contain a default export (imported as '_css').` when trying to combine styles. 

I put in a quick fix that looks like ` return <h1 css={css`${STYLES_H1};${ANCHOR};`} {...props} />;`

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -40,6 +40,7 @@ export const system = {
   black: "#2D2926",
   pitchBlack: "#0c0c0c",
   brand: "#2935ff",
+  link: "#0074D9",
   green: "#28a745",
   yellow: " #FFC940",
   red: "#ff0000",
--- a/components/system/components/Typography.js
+++ b/components/system/components/Typography.js
@@ -3,6 +3,21 @@ import * as Constants from "~/common/constants";
 
 import { css } from "@emotion/react";
 
+const ANCHOR = `
+  a {
+    font-family: ${Constants.font.text};
+    font-weight: 400;
+    text-decoration: none;
+    color: ${Constants.system.link};
+  }
+
+  a:hover {
+    text-decoration: underline;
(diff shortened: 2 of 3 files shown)


components/system/components/Typography.js
```diff
@@ -180,12 +163,23 @@ const STYLES_LI = css`
...
-    font-weight: 400;
-  }
 `;
 
 export const LI = (props) => {
-  return <li css={STYLES_LI} {...props} />;
+  return <li css={css`${STYLES_LI};${ANCHOR};`} {...props} />;
 };
+
+const ANCHOR = css`
```
jimmylee, line 172: Interesting idea, but I would prefer this instead:

```js
const ANCHOR = `
  a  {
    font-family: ${Constants.font.text};
    font-weight: 400;
    text-decoration: none;
    color: ${Constants.system.link};
  }

  a:hover {
    text-decoration: underline;
    color: ${Constants.system.link};
  }
`
```

That way you can do:

```js
const STYLES_P = css`
  box-sizing: border-box;
  font-family: ${Constants.font.text};
  font-size: ${Constants.typescale.lvl1};

  ${ANCHOR}
`;
```

And then we only need to:

```js
return <p css={STYLES_P} {...props} />
```

common/constants.js
```diff
@@ -40,6 +40,7 @@ export const system = {
   black: "#2D2926",
   pitchBlack: "#0c0c0c",
   brand: "#2935ff",
+  link: "#0074D9",
```
jimmylee, line 43: Nice color choice! I like this as a compliment to our brand blue, down to try it out and see how I feel.

PR #118, opened 2020-07-25
feat: make modal accessible

akuokojnr:
I noticed the following accessibility issues with the current modal: 
- Keyboard navigation with `Tab`, `Shitt` + `Tab` and `Esc` was not working.
- The current modal does not have label to give context to users with assistive technology such screen users.

This PR fixes these issues and it follows the [guidelines](https://www.w3.org/TR/wai-aria-practices-1.2/#dialog_modal) by the W3C.

### DIFF
--- a/components/system/components/GlobalModal.js
+++ b/components/system/components/GlobalModal.js
@@ -1,4 +1,5 @@
 import * as React from "react";
+import FocusLock from "react-focus-lock";
 import * as Constants from "~/common/constants";
 import * as SVG from "~/components/system/svg";
 import * as Strings from "~/common/strings";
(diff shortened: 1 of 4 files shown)


components/system/components/GlobalModal.js
```diff
@@ -1,4 +1,5 @@
 import * as React from "react";
+import FocusLock from "react-focus-lock";
```
jimmylee, line 2: Nit: @akuokojnr can you put this import in better order on the imports. Add a new line, and put it underneath the `*` imports.

components/system/components/GlobalModal.js
```diff
@@ -61,21 +64,47 @@ export class GlobalModal extends React.Component {
     this.setState({ modal: null });
   };
 
+  _handleDocumentKeydown = (e) => {
+    if (this.state.modal && e.keyCode === 27) {
+      this.setState({ modal: null });
```
jimmylee, line 69: put the `e.stopPropagation()` inside of the if statement.

jimmylee:
Thanks for doing this :)

PR #120, opened 2020-07-28
Slider component

martinalong:
- Slider component with options for continuous/discrete, input box, bubble
- Minor fix for settings component regex to make sure only integers are accepted

image
image


components/system/components/Slider.js
```diff
@@ -0,0 +1,242 @@
...
+
+const STYLES_BAR_CONTAINER = css`
+  height: 48px;
+  display: grid;
+  grid-template-columns: 1fr auto;
+  align-items: center;
+`;
+
+const STYLES_SLIDER_BAR = css`
+  width: calc(100% - 24px);
```
jimmylee, line 18: Why do you need to subtract 24px?

components/system/components/Slider.js
```diff
@@ -0,0 +1,242 @@
...
+  top: 10px;
+  left: 12px;
+  box-sizing: border-box;
+`;
+
+const STYLES_ACTIVE_SLIDER_BAR = css`
+  height: 4px;
+  border-radius: 2px 0px 0px 2px;
+  background-color: ${Constants.system.brand};
+  position: absolute;
```
jimmylee, line 32: Does this need to be three absolute elements? Or just two?

components/system/components/Slider.js
```diff
@@ -0,0 +1,242 @@
...
+  toValue = (px) => {
+    return px * this.state.conversion + this.props.min;
+  };
+
+  /* Converts from return value to px width */
+  toPx = (value) => {
+    return (value - this.props.min) / this.state.conversion;
+  };
+
+  /* NOTE (Martina): make sure you only query this.state.value and ui.deltaX once and save the value. Querying
```
jimmylee, line 141: // NOTE(Martina):
// Keep comment style the same throughout the codebase.

components/system/components/Slider.js
```diff
@@ -0,0 +1,242 @@
...
+    this.setState({ value: px, input: e.target.value });
+  };
+
+  _handleBlur = (e) => {
+    this.setState({ input: this.props.value });
+  };
+
+  render() {
+    return (
+      <div>
```
jimmylee, line 172: You don't need this parent `div`, replace with `<React.Fragment>`

components/system/components/Slider.js
jimmylee: Imagine this can be done with one relative element, and two absolute elements. Bonus points if you do it with only one absolute element.

components/system/components/Slider.js
martinalong: I have it so when the handle is at the far left or right, the bar end is in the middle of the dot. That way it can fully reach all the values along the whole slider. If I had the bar extend the whole length, measurements based off of the width of the bar (which are used to calculate the returned `value`) would be off.

jimmylee: Interesting, clever idea! I'm going to do a scratch to see if I can simplify this and show you at a later point.

PR #125, opened 2020-07-29
Added country markers to globe

martinalong:
image


jimmylee:
NICE ONE @martinalong !!

PR #137, opened 2020-08-01
Friends list 

martinalong:
- Friends list component
- Refactored button "full" to be a prop rather than a separate component (left the option to use <ButtonPrimaryFull /> since some people may already be using it, but updated the documentation to reflect the other format instead so we can be consistent across our components)
- Correction to settings component to remove "auto approve deals" setting since it isn't a part of powergate
- Correction to documentation for storage deals component so it is compatible with async fetching
- Edited Powergate Host URL to retrieve current value from node_common/constants.js instead of hard coding it. This way we no longer need to change each example everytime the host URL changes

Up next for friends features:
- Seeing storage/retrieval deals from friends

image

### DIFF
--- a/components/system/components/fragments/TableComponents.js
+++ b/components/system/components/fragments/TableComponents.js
@@ -100,6 +100,7 @@ const STYLES_COLUMN = css`
   align-items: flex-start;
   justify-content: space-between;
   align-self: stretch;
+  min-width: 10%;
 `;
 
 const STYLES_TOP_COLUMN = css`
@@ -109,6 +110,7 @@ const STYLES_TOP_COLUMN = css`
   justify-content: space-between;
   align-self: stretch;
   transition: 200ms ease all;
+  min-width: 10%;
 `;
 
 const STYLES_CONTENT = css`
@@ -120,6 +122,7 @@ const STYLES_CONTENT = css`
   word-break: break-word;
   overflow-wrap: anywhere;
   font-size: 12px;
+  min-width: 10%;
 `;
 
 const STYLES_CONTENT_BUTTON = css`
--- a/components/system/components/fragments/TableComponents.js
+++ b/components/system/components/fragments/TableComponents.js
@@ -54,27 +54,23 @@ const STYLES_TABLE_TAG = css`
   white-space: nowrap;
 `;
 
+const Tag = (props) => {
+  return <span css={STYLES_TABLE_TAG} {...props} />;
+};
+
 const COMPONENTS_DEAL_DIRECTION = {
-  "1": (
-    <span css={STYLES_TABLE_TAG} style={{ background: Constants.system.green }}>
-      storage
-    </span>
-  ),
-  "2": <span css={STYLES_TABLE_TAG}>retrieval</span>,
+  "1": <Tag style={{ background: Constants.system.green }}>storage</Tag>,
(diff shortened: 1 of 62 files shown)


components/system/components/fragments/TableComponents.js
```diff
@@ -100,7 +100,6 @@ const STYLES_COLUMN = css`
   align-items: flex-start;
   justify-content: space-between;
   align-self: stretch;
-  min-width: 10%;
```
jimmylee, line 103: `min-width: 10%` is a style added to catch edge cases on safari and mobile safari, did you test in those browsers?

components/system/modules/FriendsList.js
```diff
@@ -0,0 +1,245 @@
...
+  ${STYLES_BUTTON}
+  color: ${Constants.system.brand};
+`;
+
+const STYLES_REJECT_BUTTON = css`
+  ${STYLES_BUTTON}
+  color: ${Constants.system.black};
+`;
+
+const centerLeftStyle = {
```
jimmylee, line 38: Don't break the naming trend for these. use `STYLES_INLINE_`

components/system/modules/FriendsList.js
```diff
@@ -0,0 +1,245 @@
...
+                style: centerRightStyle,
+                contentStyle: { padding: "0px" },
+              },
+            ],
+            rows: this.props.data.requests.map((each) => {
+              return {
+                id: each.user,
+                user: (
+                  <a
+                    href={"/" + each.user}
```
jimmylee, line 101: {\`/${each.user}\`}

components/system/modules/FriendsList.js
```diff
@@ -0,0 +1,245 @@
...
+                      src={each.img || genericImg}
+                      alt=""
+                      style={{ width: "40px" }}
+                    />
+                  </div>
+                ),
+                children: (
+                  <div>
+                    <div
+                      style={{
```
jimmylee, line 200: Why isn't this just a class?

pages/experiences/filecoin-settings.js
```diff
@@ -52,13 +54,9 @@ const EXAMPLE_CODE = `import * as React from "react";
 import { FilecoinSettings } from "slate-react-system";
 import { createPow } from "@textile/powergate-client";
 
-const PowerGate = createPow({ host: "http://pow.slate.textile.io:6002" });
+const PowerGate = createPow({ host: "${POWERGATE_HOST}" });
```
jimmylee, line 57: People don't have access to `POWERGATE_HOST` outside of Slate, so please change this back.

components/system/modules/FriendsList.js
```diff
@@ -0,0 +1,245 @@
...
+
+  _handleClick = (e) => {
+    this.setState({ selectedRowId: e.target.value });
+  };
+
+  render() {
+    return (
+      <React.Fragment>
+        <div
+          style={{ fontSize: Constants.typescale.lvl2, marginBottom: "8px" }}
```
jimmylee, line 64: Just make these classes? It will be faster than computing inline styles performance wise.

components/system/components/fragments/TableComponents.js
martinalong: Ahh good to know. No worries, I can leave it in there!

PR #140, opened 2020-08-02
Button full syntax standardization

martinalong:
- Converting `<ButtonPrimaryFull />` to `<ButtonPrimary full />`
- Still has support for old syntax, in case others may be using it

### DIFF
--- a/components/system/components/Buttons.js
+++ b/components/system/components/Buttons.js
@@ -59,7 +59,58 @@ const STYLES_BUTTON_PRIMARY = css`
   }
 `;
 
+const STYLES_BUTTON_PRIMARY_FULL = css`
+  ${STYLES_BUTTON_FULL}
+  box-shadow: 0 1px 4px rgba(0, 0, 0, 0.15);
+  cursor: pointer;
+  background-color: ${Constants.system.brand};
+  color: ${Constants.system.white};
+
+  :hover {
+    background-color: #003fe3;
+  }
+
+  :focus {
+    box-shadow: inset 0 0 5px 2px rgba(0, 0, 0, 0.3);
+    background-color: ${Constants.system.brand};
+    outline: 0;
+    border: 0;
+  }
+`;
+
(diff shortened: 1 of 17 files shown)


components/system/components/Buttons.js
```diff
@@ -59,7 +59,58 @@ const STYLES_BUTTON_PRIMARY = css`
...
+  :focus {
+    box-shadow: inset 0 0 5px 2px rgba(0, 0, 0, 0.3);
+    background-color: ${Constants.system.brand};
+    outline: 0;
+    border: 0;
+  }
+`;
+
 export const ButtonPrimary = (props) => {
+  if (props.full) {
```
jimmylee, line 82: The only difference between `props.full` and `!props.full` is some CSS, this will actually produce some bugs because `loading` is not exclusive to just full buttons. What I would do is just another class instead.

components/system/components/Buttons.js
```diff
@@ -59,7 +59,58 @@ const STYLES_BUTTON_PRIMARY = css`
...
+          children={props.children}
+          type={props.label}
+          htmlFor={props.htmlFor}
+        />
+      );
+    }
+
+    return (
+      <button
+        css={STYLES_BUTTON_PRIMARY_FULL}
```
jimmylee, line 106: This is incorrect because its not the full option, but refer to my previous comment regardless.

components/system/components/Buttons.js
```diff
@@ -241,6 +294,15 @@ const STYLES_BUTTON_DISABLED_FULL = css`
   }
 `;
 
+export const ButtonDisabled = (props) => {
+  return (
+    <button
+      css={props.full ? STYLES_BUTTON_DISABLED_FULL : STYLES_BUTTON_DISABLED}
```
jimmylee, line 300: We should follow this structure for the `ButtonPrimary` and `ButtonSecondary`.

Issue #129, opened 2020-07-29
Feature: Offline Client: MacOS


jimmylee:
## Purpose

We want to have an offline client ready for Slate at launch.

### Architecture

Previously I made @uonai jump through some hoops to run a local server and use NextJS for builds.

* **That is no longer the case**. I have made some changes that greatly simplified the entry point for the application.

```js
import * as React from "react";

import Application from "~/components/core/Application";

export const getServerSideProps = async (context) => {
  return {
    props: { ...context.query },
  };
};

export default class ApplicationPage extends React.Component {
  render() {
    return <Application viewer={this.props.viewer} />;
  }
}
```

* A single import of `Application` means that the whole application is portable and usable in many contexts.
* We should make a new electron entry point file that populates viewer the same way (if it exists).
* **Viewer doesn't need to exist** to load the application, so the application should load the sign in screen.
* We need to modify https://github.com/filecoin-project/slate/blob/main/common/actions.js accordingly to support calls against `https://slate.host`.
* We need to also add code to use web storage instead of `cookies` in the electron context.

### Deliverable

* An electron application with a new entry point that makes HTTP calls against `https://slate.host`

### Notes

This sets the stage well for offline Slate, because with the new entry point we can add properties like:

```js
export default class ApplicationElectronContext extends React.Component {
  render() {
    return <Application isOffline localFiles={this.props.localFiles} viewer={this.props.viewer} />;
  }
}
```

Let me know if you have any questions!!

jimmylee:
I hate to continue punting this but after I finish the Slate Page (Public) / Slate Page (Private) concept I'll be taking this task over.

Justin2997:
@jimmylee  I can try to do this !

PR #143, opened 2020-08-03
Colin/electron updates

uonai:
Please see updates for #129  :
1. Tracking whether client is online / offline. We can reference this value for maintaining local file storage.
2. Separating electron prod from dev.
3. Redirect to /application page in electron build.

A couple notes after testing: 
- Cookies seem to be working fine in electron, I'm not sure if it makes sense to switch over to web storage unless there are requirements down the line that will require this. 
- There aren't any problems executing against a local node server. We can reference https://slate.host but i'm concerned that decoupling the API layer from the electron + nextjs layer can cause issues down the line if these interfaces become out of sync. Spinning up the API layer and the electron layer locally seems like the path of least resistance unless you see this as something that should be phased out.
- We are _always_ requiring that the user login every time the application refreshes. With this flow, if the user doesn't have an internet connection, then the application isn't usable. Ideally we'd persist this login session in a different way (possibly in a localdb?) to make this application usable offline.


components/core/Application.js
```diff
@@ -64,6 +68,11 @@ export default class ApplicationPage extends React.Component {
     window.removeEventListener("drop", this._handleDrop);
   }
 
+  _handleOnlineStatus = async () => {
+    window.alert(navigator.onLine ? "online" : "offline");
```
jimmylee, line 72: Nit: Any reason for `onLine` versus just `online`? seems like an interesting distinction haha.

uonai:
Pulling onLine from the Mozilla JS docs: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine

PR #144, opened 2020-08-05
AvatarGroup Component

jasonleyser:
Created an AvatarGroup component

### DIFF
--- a/components/system/components/AvatarGroup.js
+++ b/components/system/components/AvatarGroup.js
@@ -1,0 +1,77 @@
+import * as React from "react";
+import * as Constants from "~/common/constants";
+
+import { css } from "@emotion/react";
+
+const STYLES_UL = css`
+  list-style-type: none;
+
+  :hover li:nth-child(n + 2) {
+    margin-left: 0px;
+  }
+`;
+
+const STYLES_LI = css`
+  border-radius: 50%;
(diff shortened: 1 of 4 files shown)


components/system/components/AvatarGroup.js
```diff
@@ -0,0 +1,77 @@
+import * as React from "react";
+import * as Constants from "~/common/constants";
+
+import { css } from "@emotion/react";
+
+const STYLES_UL = css`
+  list-style-type: none;
+
+  :hover li:nth-child(n + 2) {
```
jimmylee, line 9: What do you need this for?

jasonleyser: do you mean the `list-style-type`?

jimmylee: No I mean this :hover li:nth-child(n + 2) {}

Why isn't the hover rule inside of the `li` element?

jasonleyser: the transition animation doesn't work when I move it to the `li` element

jimmylee: Gotcha, we can refine this as we use it. Thanks ^.^

jasonleyser: sounds good!

PR #146, opened 2020-08-05
Updated CSS Animation names

jasonleyser:
https://github.com/filecoin-project/slate/issues/134


jimmylee:
Thanks a bunch, this is a huge help in case we bring in another animation library we won't have any possibility for collisions :)

PR #153, opened 2020-08-12
Global tooltip component

martinalong:
- Global tooltip component, auto repositioning to be visible in the window. Option to specify orientation
     - GlobalTooltip: renders the tooltips at the root level
     - TooltipWrapper: wraps the tooltip anchor to determine its position and size
     - TooltipAnchor: out of the box tooltip anchor that renders a string

image


jimmylee:
Much Better :)

PR #157, opened 2020-08-17
Analytics API integration

uonai:
There are 6 endpoints that return meaningful data: blocks, blockParents, blockRewards, blockMessages, receipts, message.

API Project: https://github.com/filecoin-project/slate-stats
Example API call: https://slate-stats-dev.azurewebsites.net/blocks

All endpoints are returning 100 rows at the moment to preserve data, we can add pagination as needed (I think 100 rows per page sounds like a reasonable amount of data).


jimmylee:
LGTM! This is exactly what I wanted, thank you.

PR #162, opened 2020-08-19
fix(hub): update hub cli lib and use new crypto ident class

andrewxhill:
- updates to latest hub cli with internal endpoint update
- replaces libp2p crypto with the newer privatekey class. stored strings should remain compatible


jimmylee:
LGTM, will try this now.

node_common/utilities.js
```diff
@@ -6,8 +6,7 @@ import * as Constants from "~/node_common/constants";
 import JWT from "jsonwebtoken";
 import BCrypt from "bcrypt";
 
-import { Buckets } from "@textile/hub";
-import { Libp2pCryptoIdentity } from "@textile/threads-core";
+import { Buckets, PrivateKey } from "@textile/hub";
```
andrewxhill, line 9: PrivateKey uses the same internal key types, the class has some adding signing, encryption, decryption methods for use later

package.json
```diff
@@ -43,9 +43,8 @@
     "@emotion/css": "11.0.0-next.12",
     "@emotion/react": "11.0.0-next.12",
     "@emotion/server": "11.0.0-next.12",
-    "@textile/hub": "^0.5.2",
+    "@textile/hub": "^0.5.3",
```
andrewxhill, line 46: internal host path fix for new hub

PR #188, opened 2020-08-28
Added data randomizer, sort fixes for functions

jordattebayo:
Added randomly generated dates and values, exposed a bug in maxX and minX functions. Fixed bugs! 

Not much but feels like a lot!


jimmylee:
Nice incremental work! LGTM

PR #207, opened 2020-09-04
Styling updates

gndclouds:
Trying to PR more 😅 . Style changes and Draft of the Community Pages + Updates to Download Page + Changed Email on Privacy page. 

Know Todos (for my morning):
- Final Assets
- Copy Check
- Mobile Padding on Comunity and Download


pages/_/marketing-candidate/download.js
```diff
@@ -0,0 +1,225 @@
...
+
+import { css } from "@emotion/react";
+
+import WebsitePrototypeWrapper from "~/components/core/WebsitePrototypeWrapper";
+import WebsitePrototypeHeader from "~/components/core/NewWebsitePrototypeHeader";
+import WebsitePrototypeFooter from "~/components/core/NewWebsitePrototypeFooter";
+
+const STYLES_ROOT = css`
+  padding: 16px 88px;
+  section {
```
jimmylee, line 14: There is a ton of potential for side effects here. Please do not do this anymore (styling ANY of the tags like this)

I've remarked about this to @tarafanlin as well. I'm going to let it slide for the sake of progress but will make you both come back and clean this up right after V1 launch.

jimmylee (approved):
LGTM, with a caveat that I don't want to see `h1`, `button`, `section`, `h2, `h3`, `p`, `img`, styled this way and I've mentioned to you and @tarafanlin before so please hold each other accountable to this so we can maintain code quality.

* None of the design system components we have are immutable, each `STYLES_*` block is a guarantee that the styles are scoped to a component. We make exceptions within reason, but the ratio out of 100 choices is 5/95 when we break this rule. This is why we never experience side effects inside of the application.

* There are performance issues to nested tag styles. I can provide literature or you can investigate when you have the time. Whats the gist? O(1) lookup for classNames is what we are aiming for.

pages/_/marketing-candidate/download.js
gndclouds: Will refactor.

PR #218, opened 2020-09-06
@gndclouds/community page small devices

gndclouds:
Working on 
- Pokemon style card flair
- Making the get involved section more compelling as discussed on the call. (Some logic was tripping me up, I wanted to make a PR before I broke it :-) )


gndclouds:
Oh! And I'm playing with some blender avatars ;-)

jimmylee (approved):
LGTM, there are nits but we need to move on this so thank you for doing it :)

PR #224, opened 2020-09-07
@gndclouds/general-benchmark-improvements Updated all assets to new compressed versions

gndclouds:
What changed:
- Images and videos are about 50% smaller, 
- Added images previews to example slates. 

Would love feedback on if there is a better way to implement lines **412** to **510**. It feels like, I could have wrapped each card element in one more `div` to make the CSS classes less repetitive but not sure if one way is better for performance over the other. 🤷🏻‍♂️ LMK!


gndclouds:
Also, the [Slate/Marketing](https://slate.host/slate/marketing) is a little messy, I can remove the duplicates once this is merged and pushed to production. (Didn't want to anything currently being used) Currently, updated files are marked with _-small_ .

pages/index.js
```diff
@@ -761,111 +902,168 @@ export default class IndexPage extends React.Component {
             <br />
             <div css={STYLES_SLATE_CARD_GROUP}>
               <div css={STYLES_SLATE_CARD}>
-                <a css={STYLES_SLATE_CARD_PARAGRAPH} href="https://slate.host/gndclouds/urban-gardens" target="_blank">
-                  <div css={STYLES_SLATE_CARD_TEXT}>
-                    <div css={STYLES_SLATE_CARD_TITLE}>Urban gardens</div>
-                    <div css={STYLES_SLATE_CARD_EXPLAINER}>
-                      <div css={STYLES_SLATE_CARD_PARAGRAPH}>@gndclouds</div>
-                      <div css={STYLES_SLATE_CARD_PARAGRAPH}>-&gt;</div>
+                <div css={STYLES_SLATE_CARD_URBAN_GARDENS}>
```
jimmylee, line 905: There is a lot of code duplication here, we may want to create a:

```
SOME_CONSTANT_SET.map(each => {
   return (<div>Hello World</div>);
});
```

jimmylee (changes requested):

gndclouds:
almost have the array.map working.

PR #226, opened 2020-09-07
@gndclouds/nav colors

gndclouds:
Still need to address the white burger on the community page issue.  I was having some trouble passing props into a non-class component. But this update makes the nav more readable when over video and image elements.


jimmylee (approved):
LGTM

PR #229, opened 2020-09-08
Rebased and ready for PR

gndclouds:
ok, let's try this one!

- There is a new HoverTileColorful in the style of the pokemon card style. Which is almost complete! And tablet padding.


jimmylee:
LGTM! We'll get to the nits outside of this PR thread.

PR #237, opened 2020-09-09
[WIP] Textile Staging

jimmylee:
This PR will land a change that will make...

**http://localhost:1337** -> Its customizable where you point things. Right now without this env variable:

```js
TEXTILE_HUB_STAGING_HOST=https://webapi.hub.staging.textile.io
```

It will point at production, but with the env variable, it will point at whatever URL is set.

**https://filecoin.onrender.com** -> Always pointed to Textile Staging
**https://slate.host** -> Always pointed at Textile production.

### DIFF
--- a/pages/api/users/delete.js
+++ b/pages/api/users/delete.js
@@ -39,7 +39,16 @@ export default async (req, res) => {
   await Data.deleteSlatesForUserId({ userId: user.id });
 
   const i = await PrivateKey.fromString(user.data.tokens.api);
-  const b = await Buckets.withKeyInfo(TEXTILE_KEY_INFO);
+
+  let b;
+  if (!Strings.isEmpty(Environment.TEXTILE_HUB_STAGING_HOST)) {
+    b = await Buckets.withKeyInfo(TEXTILE_KEY_INFO, {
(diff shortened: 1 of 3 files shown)


pages/api/users/delete.js
```diff
@@ -39,7 +39,16 @@ export default async (req, res) => {
...
+
+  let b;
+  if (!Strings.isEmpty(Environment.TEXTILE_HUB_STAGING_HOST)) {
+    b = await Buckets.withKeyInfo(TEXTILE_KEY_INFO, {
+      host: Environment.TEXTILE_HUB_STAGING_HOST,
+    });
+  } else {
+    b = await Buckets.withKeyInfo(TEXTILE_KEY_INFO);
+  }
+
```
andrewxhill, line 51: maybe slightly tighter. just,

```
const options = {}
if (!Strings.isEmpty(Environment.TEXTILE_HUB_STAGING_HOST)) {
  options.host = Environment.TEXTILE_HUB_STAGING_HOST
}
const b = await Buckets.withKeyInfo(TEXTILE_KEY_INFO, options);
```

jimmylee: way better thanks

PR #266, opened 2020-09-12
Refactored for Chart to be single component. Style updates.

jordattebayo:
Incremental improvements - style updates, refactored createChart function to delete out of base.


pages/_/stats.js
```diff
@@ -15,22 +15,22 @@ const randomDate = (start, end) => {
...
-  { id: 8, date: randomDate(new Date(2018, 0, 1), new Date()), category: "2", value: getRandomInt(1000) },
-  { id: 9, date: randomDate(new Date(2018, 0, 1), new Date()), category: "2", value: getRandomInt(1000) },
-  { id: 10, date: randomDate(new Date(2018, 0, 1), new Date()), category: "2", value: getRandomInt(1000) },
-  { id: 11, date: randomDate(new Date(2018, 0, 1), new Date()), category: "3", value: getRandomInt(1000) },
-  { id: 12, date: randomDate(new Date(2018, 0, 1), new Date()), category: "3", value: getRandomInt(1000) },
-  { id: 13, date: randomDate(new Date(2018, 0, 1), new Date()), category: "3", value: getRandomInt(1000) },
-  { id: 14, date: randomDate(new Date(2018, 0, 1), new Date()), category: "3", value: getRandomInt(1000) },
-];
+  let data = [
+    { id: 1, date: randomDate(new Date(2018, 0, 1), new Date()), category: "1", value: getRandomInt(1000) },
```
jimmylee, line 19: I think these dates should represent something linear right? I don't think this makes sense otherwise.

Jan 1, Jan, 2, Jan 3

for example

components/stats/Chart.js
```diff
@@ -204,26 +267,124 @@ export default class Chart extends React.Component {
...
+          <text css={STYLES_CHART_TEXT} textAnchor="middle" x={tick.x} y="575">
+            {`${month} ${year}`}
+          </text>
+        </g>
       );
-    } else {
-      return <React.Fragment></React.Fragment>;
+      tickLines.push(tickLine);
     }
+    ReactDOM.render(tickLines, document.getElementById("tickContainer"));
```
jimmylee, line 366: You don't need `ReactDOM`, You can just use `this.state.ticks.map(each => {})` to achieve your goals and declare that inside of `render` I believe.

components/stats/Chart.js
```diff
@@ -204,26 +267,124 @@ export default class Chart extends React.Component {
...
-            ticks={this.state.ticks}
-            xLabel={this.state.xLabel}
-            yLabel={this.state.yLabel}
-            organizedData={this.state.organizedData}
-          />
-        </React.Fragment>
+        <circle key={index} cx={g.x} cy={g.y} r="2" css={STYLES_CHART_CIRCLE} />
+      );
+    });
+    ReactDOM.render(allCircles, document.getElementById("circles"));
```
jimmylee, line 302: `ReactDOM` isn't necessary, you should be able to get this all within `render` using `map` directly inside of a render function. For example, this is acceptable:

```
render() {
   return (
      <div>{allCircles}</div>
   )
}
```

jimmylee (changes requested):
Making progress! I like it.

pages/_/stats.js
jordattebayo: The dates are sorted before they are displayed so they are always linear. The randomDate function is a means to test the robustness of the component by throwing a variety of dates. I can revert it back to the linear hard coded sample data I had originally if you would prefer!

components/stats/Chart.js
jordattebayo: It took some work but I refactored to remove reactDOM from the component and used your suggested method!

jordattebayo:
I think I've addressed your concerns, so please let me know if there is anything additional I need to update or change! Or if I should create a new PR.

PR #268, opened 2020-09-13
Marketing pages: misc styling + remake table for download page

tarafanlin:
making this PR so Will can get the updated version and fix the safari bug on top of it.

### DIFF
--- a/pages/_/marketing-candidate/download.js
+++ b/pages/_/marketing-candidate/download.js
@@ -1,7 +1,6 @@
 import * as React from "react";
 import * as Constants from "~/common/constants";
 import * as Actions from "~/common/actions";
-import * as System from "~/components/system";
 
 import { css } from "@emotion/react";
 
@@ -17,11 +16,13 @@ const STYLES_ROOT = css`
 
 const STYLES_H1 = css`
   font-size: ${Constants.typescale.lvl5};
+  font-weight: 700;
   line-height: 1.3;
   padding: 0px 0px 24px 0px;
   letter-spacing: -0.021rem;
   width: 100%;
   color: ${Constants.system.slate};
+
   @media (max-width: ${Constants.sizes.tablet}px) {
     font-size: ${Constants.typescale.lvl5};
     padding: 0px 0px 16px 0px;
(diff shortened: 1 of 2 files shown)


pages/_/marketing-candidate/download.js
```diff
@@ -17,11 +16,13 @@ const STYLES_ROOT = css`
 
 const STYLES_H1 = css`
   font-size: ${Constants.typescale.lvl5};
+  font-weight: 700;
```
jimmylee, line 19: Take font-weight out, again use the font-family.

pages/_/marketing-candidate/download.js
```diff
@@ -48,17 +50,18 @@ const STYLES_H3 = css`
...
   justify-content: space-between;
   max-width: 56vw;
   width: 100%;
   align-self: center;
+  margin-bottom: 16px;
 
   @media (max-width: ${Constants.sizes.mobile}px) {
     width: 100%;
-    right: 24px;
+    max-width: 100vw;
```
jimmylee, line 64: What does this do?

pages/index.js
```diff
@@ -165,6 +138,7 @@ const STYLES_LINK = css`
 
 const STYLES_H1 = css`
   font-size: ${Constants.typescale.lvl5};
+  font-weight: 700;
```
jimmylee, line 141: Don't use font-weight, use font-family and add the fonts you need.

pages/index.js
```diff
@@ -183,6 +157,7 @@ const STYLES_H1 = css`
 
 const STYLES_H2 = css`
   font-size: ${Constants.typescale.lvl3};
+  font-weight: 500;
```
jimmylee, line 160: Don't use font-weight unless you're using a system font that has the weightings, we use inter and use a specific woff per font-weight.

pages/index.js
```diff
@@ -201,6 +176,7 @@ const STYLES_H2 = css`
 
 const STYLES_H3 = css`
   font-size: ${Constants.typescale.lvl2};
+  font-weight: 500;
```
jimmylee, line 179: See above

jimmylee (changes requested):

pages/index.js
tarafanlin: wait I thought you wanted us to add `font-weight: 400` yesterday? Is there a difference?

pages/_/marketing-candidate/download.js
tarafanlin: to overite the `max-width: 56vw` in desktop

jimmylee (approved):

Issue #281, opened 2020-09-14
Bug: Link to Slate button creates a link to another user's slate


charlesvinette:
# Purpose

While creating my first Slate, which I called `cyber`, clicking the `Link` button in the upper right would create a link and bring me to `gndclouds / cybernetics` instead of my `cyber` slate.

# Deliverable

Clicking the `Link` button should create and bring me to my own Slate.

Image of my Slate:
<img width="1676" alt="Screen Shot 2020-09-14 at 4 26 38 PM" src="https://user-images.githubusercontent.com/7192823/93134819-4ffa5a80-f6a7-11ea-9bb7-2bc5b49271ac.png">

Image of what clicking the Link button brings me to:
<img width="1423" alt="Screen Shot 2020-09-14 at 4 26 45 PM" src="https://user-images.githubusercontent.com/7192823/93134841-58529580-f6a7-11ea-8799-a92b84c03195.png">

jimmylee:
Thank you @charlesvinette this is huge, we'll track it down.

PR #286, opened 2020-09-15
fix: deeplink slate query bug

akuokojnr:
This PR fixes issue #281.

Previously, if a user create a slate (e.g. `cyber`) which has a name similar to the name of another slate (e.g. `cybernetics`), the query returns all of them and selects the first one. 

This PR ensures that the name of deeplink query is the one actually returned.


jimmylee:
LGTM thanks :)

Issue #282, opened 2020-09-15
Bug: Account Settings bio boarder 


gndclouds:
# Purpose

Under **Account settings**, the Bio section does not have a border

![image](https://user-images.githubusercontent.com/1757261/93171349-80033580-f75b-11ea-833c-c6ddeb84a401.png)

# Deliverable

Bio text area should have a border which matches username text area

# Dependencies

N/A

jimmylee:
Thank you @charlesvinette for jumping on this and @gndclouds for reporting!

PR #292, opened 2020-09-16
Account settings Bio border fix for Safari

charlesvinette:
Little fix for #282 which doesn't currently have a border showing up around TextAreas on Safari.

### DIFF
--- a/components/system/components/Textarea.js
+++ b/components/system/components/Textarea.js
@@ -24,8 +24,7 @@ const STYLES_TEXTAREA = css`
   border: 0;
   transition: 200ms ease all;
   padding: 16px 24px 16px 24px;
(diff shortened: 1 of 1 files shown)


jimmylee:
@charlesvinette thank you for diving in! Can you try an alternate strategy here

which was:

```
will-change: transform;
-webkit-appearance: none;
```

Because we have auto-prefixing so I think the problem is getting the compositor engine to work on Safari.

charlesvinette:
@jimmylee Weirdly enough, adding `will-change: transform;` didn't fix it, but re-ordering the properties of `box-shadow` did the trick. Was able to remove the web-kit specific css.

jimmylee:
Awesome!! Thank you!

PR #316, opened 2020-09-22
fix(api): directly use thread client to generate bucket thread if not exit

andrewxhill:


node_common/utilities.js
```diff
@@ -132,7 +132,29 @@ export const getBucketAPIFromUserToken = async (token, user) => {
...
+      const res = await client.getThread('buckets')
+      buckets.withThread(res.id.toString())
+    } catch (error) {
+      if (error.message !== 'Thread not found') {
+        throw new Error(error.message)
+      }
+      const newId = ThreadID.fromRandom()
+      await client.newDB(newId, 'buckets')
+      threadID = newId.toString()
+      buckets.withThread(threadID)
```
carsonfarmer, line 149: I'd also like to get an idea of the actual thread id that is being used here. I saw some issues in the past with funny behavior around Buffer in thread-id. Can we log the id or otherwise debug what's actually making it to the bucket's context for that field? Even dumping the whole bucket context would be useful here.

PR #329, opened 2020-09-25
encrypted buckets: encrypted buckets in production

jimmylee:


jimmylee:
* Single bucket named `encrypted-deal`
* Will effect all users, all former users using `deal` will be transitioned over automatically.

PR #333, opened 2020-09-27
changed default avatar for new users

toastts:


jimmylee:
Nice and easy ✨  LGTM

PR #334, opened 2020-09-27
reduced avatar size to double digit KB range

toastts:
Scaled the default avatar image from 1080p down to 480p so it should be a more reasonable size now.


jimmylee:
LGTM

PR #336, opened 2020-09-28
landing v.2 iteration + terms & guideline pages v.2

tarafanlin:
- restructure terms & guideline pages
- adding page content menu
- responsive for mobile & tablet


pages/terms-new.js
```diff
@@ -0,0 +1,368 @@
...
+  }
+  return (
+    <div css={STYLES_COPY_EMAIL}>
+      <a onClick={copyToClipboard}>abuse@filecoin.io</a>
+      {copySuccess}
+    </div>
+  );
+}
+
+export default class GuidelinesPage extends React.Component {
```
jimmylee, line 131: Nit Bad copy & paste?

pages/guidelines-new.js
```diff
@@ -0,0 +1,383 @@
...
+  :visited {
+    color: ${Constants.system.slate};
+  }
+`;
+
+const STYLES_COPY_EMAIL = css`
+  display: inline;
+`;
+
+export const getServerSideProps = async (context) => {
```
jimmylee, line 112: Why do you need `getServerSideProps`? is anything coming from the server?

jimmylee (changes requested):

pages/terms-new.js
tarafanlin: yep  😛

pages/guidelines-new.js
tarafanlin: actually I'm not sure,,,,,I saw other pages have this so just put it there.

jimmylee: This is only necessary if you need to pass props from the server to the client. I don't think you need this for marketing pages so you can delete it.

jimmylee:
LGTM

PR #339, opened 2020-09-29
@AllenHsing3/Preview for Mov, Mp4, PDF

AllenHsing3:
[WIP]

Pdf previews render differently depending on Chrome or Firefox due to browser settings rendering the embedded file differently. It might be better to render a canvas or save a screenshot on the backend? 
Screen Shot 2020-09-29 at 12 28 53 AM

Should preview of videos autoplay on load? Or should they be a screenshot as well?

### DIFF
--- a/components/core/SlateMediaObjectPreview.js
+++ b/components/core/SlateMediaObjectPreview.js
@@ -38,6 +38,34 @@ const STYLES_TITLE = css`
   text-overflow: break-word;
 `;
 
+const STYLES_VIDEO = css`
+  margin-top: 8px;
+  overflow: hidden;
+  pointer-events: none;
+  width: 300px !important;
+  height: auto !important;
+`;
+
+const STYLES_PDF = css`
+  width: 100%;
+  height: 76.5%;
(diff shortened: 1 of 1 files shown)


jimmylee:
Hey @AllenHsing3 !! Thanks for jumping onto this, love it when people take the initiative.

When it comes to previews, we will probably want a solution that works for all cases. So I think the best trivial solution (easiest to implement) is to allow users to pick an image as a preview photo for a PDF/Video.

_But long term we may not want to go that route_, considering that on SlateApp, each slate object that you see could be expanded to be far more expressive. What I mean by that is that we could tie multiple slate objects into one, and maybe generate the preview from the first image in a grouping, etc.

**Using a large amount of video objects/iframe objects on a screen is also a performance concern**. I think in your testing this must have worked well for a number of base cases (one iframe, one video), but there are accounts that have 50+ videos we would have to turn this off for.

I think it might be better to pick up another task, since @harisbutt is in the middle of a redesign here. I appreciate the attempt regardless ⚔️

PR #350, opened 2020-10-02
@gndclouds/community v2

gndclouds:
Updates from Slack feedback


jimmylee (approved):
LGTM, I think some of the values can be finessed more but this is a mergeable PR thanks

PR #352, opened 2020-10-04
added randomized avatar function

toastts:
made an env variable called AVATAR_SLATE_ID, has slateId for https://slate.host/slate/slate-avatars, uses that Id to take random textile urls for new user avatars


toastts:
forgot to add a variable to environments.js

PR #354, opened 2020-10-04
added randomized avatar fucntion

toastts:
made AVATAR_SLATE_ID env variable that has slateId for https://slate.host/slate/slate-avatars, rand function in managers takes a slateId and returns a random element from a slate

### DIFF
--- a/node_common/managers/slate.js
+++ b/node_common/managers/slate.js
@@ -1,0 +1,15 @@
+import * as Data from "~/node_common/data";
+
+export const getRandomSlateElementURL = async ({ id, fallback }) => {
+  if (!{ id }) {
+    return { fallback };
+  }
+
+  const query = await Data.getSlateById({ id });
+
+  if (!query || query.error) {
+    return { fallback };
+  }
+
+  return query.data.objects[Math.floor(Math.random() * query.data.objects.length)].url;
+};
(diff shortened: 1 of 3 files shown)


node_common/managers/slate.js
```diff
@@ -0,0 +1,7 @@
+import * as Data from "~/node_common/data";
+
+
+export const getRandomSlateElement = async ({id}) => {
+    const query = await Data.getSlateById({id});
```
jimmylee, line 5: catch a failure case here or this will throw loud (do an empty array)

pages/api/users/create.js
```diff
@@ -68,8 +69,7 @@ export default async (req, res) => {
     salt,
     username: newUsername,
     data: {
-      photo:
-        "https://slate.textile.io/ipfs/bafkreiexygfz4e5resu66xfviokddariztq4onuai5wii5mdd7syshftca",
+        photo: SlateManager.getRandomSlateElement({id: Environment.AVATAR_SLATE_ID}),
```
jimmylee, line 72: catch a failure case here so if people don't have env variables it will use the old case

jimmylee (changes requested):
LGTM! near perfect, just make two changes for me to catch cases for people without the env variable so its safe to deploy. I'll add this env variable to our environments now.

jimmylee (approved):
LGTM, I will patch some nits in my own pass and share them with you.

PR #356, opened 2020-10-05
added simple prettier defaults

toastts:
Looked at our code and looked at [prettier](https://prettier.io/docs/en/options.html) and put the options we're using.

### DIFF
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,23 @@
+{
+   "printWidth":100,
+   "tabWidth":2,
+   "useTabs":false,
+   "semi":false,
+   "singleQuote":false,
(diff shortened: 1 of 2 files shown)


.prettierrc
```diff
@@ -0,0 +1,23 @@
+{
+   "printWidth":100,
+   "tabWidth":2,
+   "useTabs":false,
+   "semi":false,
```
jimmylee, line 5: change to true :)

jimmylee: `"semi":true,`

jimmylee (approved):
Love this, thank you @toastts !!

PR #364, opened 2020-10-08
adding 50GB storage explainer

tarafanlin:
- add explainer for 50GB storage
- fixing margin for empty data view


jimmylee (approved):
You're killing it Tara, very nice.

PR #367, opened 2020-10-10
Reproduced ERR_CONNECTION_RESET - Finally able to reproduce the bug with Buckets by throttling my Stream.

jimmylee:
In this PR, you will be able to clone and run locally and reproduce

* Bucket bricking
* Attempted solution for backpressuring in NodeJS

# Introduction

I noticed something really interesting as I was piping my `WritableStream` into a `ReadableStream` on production after adding `progress` thanks to @carsonfarmer's suggestion.

Screen Shot 2020-10-09 at 10 35 07 PM

* Almost every single time my file upload in production/local would complete to node memory way faster than `buckets.pushPath`, in fact in this example you can see that I had already completed 4GB of transfer before 370MB has been actually pushed to Textile buckets.

## So I had an idea

What if I throttled the connection so they would be at the exact same speed?

Screen Shot 2020-10-09 at 11 06 06 PM

* Eureka! This felt right, they are now operating side by side.

However, trouble in paradise:

Screen Shot 2020-10-09 at 11 08 56 PM

And another:

Screen Shot 2020-10-09 at 11 18 08 PM

* I hope this is something about rate limiting or data transfer speed
* This produces a bricked bucket every single time.

I feel like this gets us closer to what the issue is with streaming data for large files. This reproduces the bug I've been experiencing locally without having to go to production so I hope it helps us test :)

### DIFF
--- a/node_common/upload.js
+++ b/node_common/upload.js
@@ -7,6 +7,7 @@ import * as Logs from "~/node_common/script-logging";
 import AbortController from "abort-controller";
 import BusBoyConstructor from "busboy";
 import Queue from "p-queue";
+import Throttle from "~/node_common/vendor/throttle";
 
 const WORKER_NAME = "BROWSER->RENDER->TEXTILE";
 const HIGH_WATER_MARK = 1024 * 1024 * 3;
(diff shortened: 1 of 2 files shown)


jimmylee:
Found something else that someone with google storage was experiencing.

PR #371, opened 2020-10-13
[WIP] @gndclouds/my awesome scene

gndclouds:
Hey, @jimmylee here is my demo scene :-) from our tutorial with Martina.


common/navigation-data.js
```diff
@@ -80,6 +80,13 @@ export const generate = ({ library = [], slates = [] }) => [
     pageTitle: "Slates",
     children: null,
   },
+  {
+    id: "V1_NAVIGATION_WILL_SCENE",
+    decorator: "SCENE_WILL",
+    name: "Will's Scene",
+    pageTitle: "Slate",
+    children: null,
```
jimmylee, line 88: Will this work? Have you figured out how to make sure it doesn't appear on the sidebar?

jimmylee (changes requested):

common/navigation-data.js
gndclouds: Yes, it ran locally, and to avoid it running in the sidebar I just add `ignore: true,`(line 89) in the next line.

PR #372, opened 2020-10-15
@gndclouds/account setting v2

gndclouds:


gndclouds:
@cake, can you review this. I think there was one more thing you recommended I try but I can't find the notes.

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+        body: this.state.body,
+        name: this.state.name,
+        allow_filecoin_directory_listing: this.state.allow_filecoin_directory_listing,
+        allow_automatic_data_storage: this.state.allow_automatic_data_storage,
+        allow_encrypted_data_storage: this.state.allow_encrypted_data_storage,
+      },
+    });
+
+    await this.props.onRehydrate();
+    this.setState({ changingFilecoin: false });
```
jimmylee, line 183: This can all be in a single set start. They don't have to be in separate set states.

jimmylee (changes requested):
You need to patch the setState, and I need to pull this down and test, please provide a preview in your PR of the work you're doing

jimmylee:
Screen Shot 2020-10-22 at 3 07 46 PM

Deleting an account throws an exception, so make sure you test this too (Its something clientside)

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
+import * as React from "react";
+import * as System from "~/components/system";
+import * as Actions from "~/common/actions";
+import * as Strings from "~/common/strings";
+import * as Validations from "~/common/validations";
+import * as FileUtilities from "~/common/file-utilities";
+import { css } from "@emotion/react";
```
jimmylee, line 7: New line between groups

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+  left: -1px;
+`;
+
+const STYLES_COPY_INPUT = css`
+  pointer-events: none;
+  position: absolute;
+  opacity: 0;
+`;
+
+export default class SceneEditAccount extends React.Component {
```
jimmylee, line 31: `SceneEditAccountV2`

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+    changingAvatar: false,
+    changingFilecoin: false,
+    copyValue: "",
+    loading: false,
+    tab: 0,
+    contextMenu: null,
+    selectedIndex: -1,
+  };
+
+  componentDidMount = () => {
```
jimmylee, line 56: If you're going to add any event listeners, make sure when the component unmounts you do

```
componentWillUnmount
```

and

```
window.removeEventListener("keydown", this._handleDocumentKeydown)
```

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+    window.addEventListener("keydown", this._handleDocumentKeydown);
+    this.debounceInstance = this.debounce(() => {
+      if (this.state.selectedIndex !== -1) {
+        this.setState({ selectedIndex: -1 });
+      }
+      this.props.onSearch();
+    }, 500);
+  };
+
+  debounce = (fn, time) => {
```
jimmylee, line 66: debounce doesn't need to be a class method.

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+    copyValue: "",
+    loading: false,
+    tab: 0,
+    contextMenu: null,
+    selectedIndex: -1,
+  };
+
+  componentDidMount = () => {
+    window.addEventListener("keydown", this._handleDocumentKeydown);
+    this.debounceInstance = this.debounce(() => {
```
jimmylee, line 58: what does assigning this to this.debounceInstance do? you already handle the clearing of the timeout inside of your debounce function.

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+        name: this.state.name,
+      },
+    });
+
+    await this.props.onRehydrate();
+    this.setState({ changingAvatar: false, photo: url });
+  };
+
+  _handleSaveAll = async (e) => {
+    this.setState({ changingUsername: true });
```
jimmylee, line 140: Remove all of the other this.setStates and combine into one. This will cause problems.

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+    this.setState({ changingFilecoin: true });
+    this.setState({ changingBio: true });
+    this.setState({ changingUsername: true });
+    
+    if(this.state.password !== ""){
+      this.setState({ changingPassword: true });
```
jimmylee, line 146: `changingPassword: Strings.isEmpty(this.state.password) ? false : true`

scenes/SceneEditAccountV2.js
```diff
@@ -0,0 +1,411 @@
...
+    e.stopPropagation();
+    if (this.state.contextMenu === value) {
+      this._handleHide();
+    } else {
+      this.setState({ contextMenu: value });
+    }
+  };
+
+  _handleDelete = async (e, id) => {
+    this._handleHide();
```
jimmylee, line 218: This is where the error is throwing in your code, make sure you test this path.

server.js
```diff
@@ -78,9 +78,9 @@ app.prepare().then(async () => {
   server.get("/_", async (req, res) => {
     const isBucketsAvailable = await Utilities.checkTextile();
 
-    if (!isBucketsAvailable) {
-      return res.redirect("/maintenance");
-    }
+    // if (!isBucketsAvailable) {
```
jimmylee, line 81: Don't forget on a release PR to make sure that this code is put back. its important!

jimmylee (changes requested):

scenes/SceneEditAccountV2.js
gndclouds: I'll remove, originally I tried two different ways of debouncing and this was the first way I tried.

scenes/SceneEditAccountV2.js
gndclouds: Is that because the debounce is only being used once for the` _handleSaveAll`  ?

gndclouds:
Still have a few more things to tweak in the am

scenes/SceneEditAccountV2.js
jimmylee: Prefer to the other PR for my comment about `debounce` essentially you should keep it a class method but you should also store the reference to the `timer` on the class, Also not clear why they are defined differently so the code between that debounce and this debounce should be exactly the same.

jimmylee (changes requested):

PR #373, opened 2020-10-17
dividing line left of UserControls dropdown arrow

toastts:
Added a border left of the user control dropdown arrow and moved the arrow a bit to the right to center it better on desktop


toastts:
desktop
mobile

jimmylee:
Nice work :) LGTM

PR #375, opened 2020-10-20
@gndclouds/lazy loading data view

gndclouds:
Animated GIF-downsized_large

When you hit the bottom of the page the viewer gets 20 images added to it. This is done but let me know if you think anything should be implemented differently.


components/core/DataView.js
```diff
@@ -230,56 +197,55 @@ export default class DataView extends React.Component {
...
+      body.scrollHeight,
+      body.offsetHeight,
+      html.clientHeight,
+      html.scrollHeight,
+      html.offsetHeight
+    );
+    const windowBottom = windowHeight + window.pageYOffset;
+    if (windowBottom >= docHeight) {
+      this.setState({ viewLimit: this.state.viewLimit + 20 });
+    } else {
```
jimmylee, line 243: ✂️ cut this empty else?

Also might be useful to look at `common/window` and make these some utility methods we can use again.

jimmylee: You might want to add a debounce to this if you're listening to scroll, so it doesn't trigger setState an insane amount of times.

jimmylee (changes requested):
Nice work, it looks like you were cleaning up some existing stuff, one small change requested.

components/core/DataView.js
```diff
@@ -248,20 +221,44 @@ export default class DataView extends React.Component {
     window.removeEventListener("remote-data-deletion", this._handleDataDeletion);
     window.removeEventListener("remote-slate-object-remove", this._handleRemoteSlateObjectRemove);
     window.removeEventListener("remote-slate-object-add", this._handleRemoteSlateObjectAdd);
+    window.removeEventListener("scroll", this._handleCheckScroll);
```
martinalong, line 224: on line 214 you added the event listener for the function this._hanldeScroll and on this line you removed it for the function this._handleCheckScroll. Shouldn't these match? Or is this intentional?

components/core/DataView.js
```diff
@@ -248,20 +221,44 @@ export default class DataView extends React.Component {
...
-    ) {
-      this.setState({ startIndex: this.state.startIndex + VIEW_LIMIT });
-    } else if (direction < 0 && this.state.startIndex - VIEW_LIMIT >= 0) {
-      this.setState({ startIndex: this.state.startIndex - VIEW_LIMIT });
+  _handleDebounce = (func, wait) => {
+    let timeout;
+
+    return function executedFunction(...args) {
+      const later = () => {
+        clearTimeout(timeout);
```
martinalong, line 233: Not super critical, but you don't really need this `clearTimeout` here b/c the only time `later` will be called is at the end of the time out. clearTimeout is only needed if you want to cancel the timer before the time is up

components/core/DataView.js
```diff
@@ -239,6 +211,7 @@ export default class DataView extends React.Component {
       window.addEventListener("remote-update-carousel", this._handleUpdate);
     }
 
+    window.addEventListener("scroll", this._handleScroll);
```
martinalong, line 214: It looks like it's not actually using the debounce right now. B/c you have an event listener here that calls `handleScroll`, it calls `handleScroll` everytime you scroll. Instead, you want to attach the event listener to `handleCheckScroll`, which cancels the timer every time you scroll (and upon stopping scrolling, then calls handleScroll)

components/core/DataView.js
```diff
@@ -248,20 +221,44 @@ export default class DataView extends React.Component {
...
+    const html = document.documentElement;
+    const docHeight = Math.max(
+      body.scrollHeight,
+      body.offsetHeight,
+      html.clientHeight,
+      html.scrollHeight,
+      html.offsetHeight
+    );
+    const windowBottom = windowHeight + window.pageYOffset;
+    if (windowBottom >= docHeight) {
```
martinalong, line 255: Maybe play around with how many load at a time. I think 10 is too few, and feels very jerky when you're trying to scroll down repeatedly. Load at least enough to fill the height of the screen each time. (Loading will also look better with the new blur loading thing I implemented, so we won't have just grey boxes while images are waiting to load). Potentially also increase the initial starting amount to something higher than 20.

martinalong: You could also allow it to load even when you aren't at the very bottom (like maybe when windowBottom is 100px--which is approx the height of each line--away from docHeight), that would also help avoid the jump that happens when you load more in some cases.

components/core/DataView.js
```diff
@@ -228,6 +198,8 @@ export default class DataView extends React.Component {
     startIndex: 0,
```
martinalong, line 198: With your new changes, everywhere in the code where you see startIndex or this.state.startIndex you can remove it. Treat it like 0 now b/c it's always 0 (so if you see checked[this.state.startIndex + i], that turns into checked[i]). It was only necessary for when we had pagination and didn't display images starting from 0 (e.g. if you're on page 2, it would show images 20-40). Make sure you test delete and add to slate after you make this change

components/core/DataView.js
jimmylee: I really like this suggestion

gndclouds: That makes so much more sense. Then the loading can both start before they get to the true bottom and it will feel more infinite as you scroll. 🙌🏻

components/core/DataView.js
gndclouds: ok, so after implementing this in a3fbbe3 I get the following memory leak error and reload issue.

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in DataView (created by SceneHome)
    in div (created by SceneHome)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (created by ScenePage)
    in div (created by Context.Consumer)
    in EmotionCssPropInternal (created by ScenePage)
    in ScenePage (created by SceneHome)
    in SceneHome
```

gndclouds: It happens when I reach the bottom of the page, and then after all image load successfully I get kicked to the top of the page, and then this error happens. I thought it was because of the `_handleScroll` function but haven't found the issue the issue yet.

jimmylee:
Needs to resolve merge conflicts

gndclouds:
Changed to WIP until I can fix this loading bug

gndclouds:
There is a build error form main effecting this after merge d29031e 

But the error I'm working on:
1. Go to DataView
2. Scroll to the bottom of the page. (you will get kicked back up to the top)

Animated GIF-downsized_large

gndclouds:
ok! its fixed and ready!

components/core/DataView.js
```diff
@@ -240,26 +212,54 @@ export default class DataView extends React.Component {
...
-  _increment = (direction) => {
-    if (
-      direction > 0 &&
-      this.state.startIndex + VIEW_LIMIT < this.props.viewer.library[0].children.length
-    ) {
-      this.setState({ startIndex: this.state.startIndex + VIEW_LIMIT });
-    } else if (direction < 0 && this.state.startIndex - VIEW_LIMIT >= 0) {
-      this.setState({ startIndex: this.state.startIndex - VIEW_LIMIT });
+  _handleDebounce = (func, wait) => {
+    let timeout;
```
jimmylee, line 230: this reference `timeout` will get lost since its in the scope of this closure, you need to add this as a class member

```
class Component extends React.Component {
  _timeout = null;

  _handleDebounce = (callbackFn, waitMs) => {
    return function executedFunction(...args) {
      const debouncedCall = () => {
        callbackFn(...args);
      }

      clearTimeout(this._timeout);
      this._timeout = setTimeout(debouncedCall, waitMs)
    }
  }
}
```

components/core/DataView.js
```diff
@@ -240,26 +212,54 @@ export default class DataView extends React.Component {
...
+
+    return function executedFunction(...args) {
+      const later = () => {
```
jimmylee, line 233: not a great function name, should update this to something else

jimmylee (changes requested):
Patch the debounce and we should be good to go.

components/core/DataView.js
martinalong: Looks like this still isn't fixed. I saw you added an extra await this._handleCheckScroll below it, but that's different from changing this window.addEventListener("scroll", this._handleScroll); to window.addEventListener("scroll", this._handleCheckScroll);

See my slack message for more details

jimmylee:
I left some comments earlier but I think the right thing to do is to consolidate this debounce method you have in both files to a single place in `~/common/window` since they are both the same.

gndclouds:
0k makes sense!

components/core/DataViewV0.js
```diff
@@ -0,0 +1,824 @@
...
+
+const STYLES_MOBILE_HIDDEN = css`
+  @media (max-width: ${Constants.sizes.mobile}px) {
+    display: none;
+  }
+`;
+
+const delay = (ms) => new Promise((resolve) => window.setTimeout(resolve, ms));
+
+let mounted = false;
```
martinalong, line 193: You have two `mounted` variables here. You can delete this outside one and use the inside one on line 196. The way you'd call the one on line 196 is with `this._mounted`, so like `this._mounted = false`

gndclouds: sorry, this was the original file I was keeping as a backup. I can remove it now.

components/core/DataViewV0.js
```diff
@@ -0,0 +1,824 @@
...
+
+  async componentDidMount() {
+    if (!mounted) {
+      mounted = true;
+      window.addEventListener("remote-data-deletion", this._handleDataDeletion);
+      window.addEventListener("remote-slate-object-remove", this._handleRemoteSlateObjectRemove);
+      window.addEventListener("remote-slate-object-add", this._handleRemoteSlateObjectAdd);
+    }
+
+    window.addEventListener("scroll", this._handleScroll);
```
martinalong, line 215: Move this inside the `if(!mounted)` above (the purpose of that is so that it isn't called twice when there is a double mount) and change `this._handleScroll` to `this._handleCheckScroll` (otherwise it isn't calling the debounce at all)

components/core/DataView.js
```diff
@@ -592,19 +549,18 @@ export default class DataView extends React.Component {
...
                     key={each.id}
                     css={STYLES_IMAGE_BOX}
-                    onClick={() => this._handleSelect(i + this.state.startIndex)}
+                    onClick={() => this._handleSelect(i)}
                     onMouseEnter={() => this.setState({ hover: i })}
                     onMouseLeave={() => this.setState({ hover: null })}
                   >
                     <SlateMediaObjectPreview
-                      blurhash={each.blurhash}
                       url={`${Constants.gateways.ipfs}/${each.ipfs.replace("/ipfs/", "")}`}
```
martinalong, line 564: Can you add `blurhash={each.blurhash}` back into <SlateMediaObjectPreview/>? It looks like it got deleted from the line above in one of your commits. Blurhash is necessary so that the images appear as a gradient blur rather than as a grey box while loading

common/window.js
```diff
@@ -97,3 +97,16 @@ export const isMobileBrowser = (userAgent) => {
 
   return !!mobile;
 };
+
+export const _checkDebounce = (func, wait) => {
```
jimmylee, line 101: Why is this `_`? this syntax is preserved for usually a private method. Why not just call it `debounce` instead of `checkDebounce`?

components/core/DataView.js
```diff
@@ -240,26 +213,41 @@ export default class DataView extends React.Component {
...
+      html.offsetHeight
+    );
+    const windowBottom = windowHeight + window.pageYOffset;
+    if (windowBottom >= docHeight - 600) {
+      this.setState({ viewLimit: this.state.viewLimit + 30 });
     }
+    console.log(e);
   };
 
+  _handleCheckScroll = Window._checkDebounce(this._handleScroll, 200);
```
jimmylee, line 249: `Window.debounce`

jimmylee (changes requested):

gndclouds:
@jimmylee Fixed

jimmylee (approved):
LGTM

PR #376, opened 2020-10-20
image preview bubble for data list view

tarafanlin:
- show the image preview when hover over the image name
- conditioned to only work for image files

to test this go to: http://localhost:1337/_?scene=data

Screen Shot 2020-10-20 at 2 22 25 AM

### DIFF
--- a/components/core/FilePreviewBubble.js
+++ b/components/core/FilePreviewBubble.js
@@ -1,0 +1,64 @@
+import * as React from "react";
+import * as Constants from "~/common/constants";
+
+import { css, keyframes } from "@emotion/react";
+import { useState } from "react";
+import SlateMediaObjectPreview from "~/components/core/SlateMediaObjectPreview";
+import SlateMediaObject from "~/components/core/SlateMediaObject";
+
+const fadein = keyframes`
+  0%{
+    opacity: 0;
+  }
+  100%{
+    opacity: 1;
+  }
+`;
+
+const STYLES_FILE_PREVIEW_BUBBLE = css`
+  z-index: ${Constants.zindex.tooltip};
+  padding: 8px;
+  border-radius: 4px;
+  background-color: ${Constants.system.black};
+  animation: ${fadein} 200ms ease-out 1;
+  width: 200px;
+  position: absolute;
+`;
+
+const STYLES_FILE_PREVIEW = css`
+  width: 100%;
+  height: 100%;
+`;
+
+const STYLES_BUBBLE_ANCHOR = css`
+  box-sizing: border-box;
+  display: inline-flex;
+  align-item: center;
+  justify-content: center;
+  cursor: pointer;
+`;
+
+export const FilePreviewBubble = (props) => {
+  const [onHover, setHover] = useState(false);
+  const type = props.type && props.type.startsWith("image/") && onHover;
+  console.log(type);
+
+  return (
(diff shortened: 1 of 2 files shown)


components/core/FilePreviewBubble.js
```diff
@@ -0,0 +1,64 @@
+import * as React from "react";
+import * as Constants from "~/common/constants";
+
+import { css, keyframes } from "@emotion/react";
+import { useState } from "react";
+import SlateMediaObjectPreview from "~/components/core/SlateMediaObjectPreview";
```
jimmylee, line 6: Nit: new line, fix this after we commit it.

components/core/FilePreviewBubble.js
```diff
@@ -0,0 +1,64 @@
...
+    opacity: 1;
+  }
+`;
+
+const STYLES_FILE_PREVIEW_BUBBLE = css`
+  z-index: ${Constants.zindex.tooltip};
+  padding: 8px;
+  border-radius: 4px;
+  background-color: ${Constants.system.black};
+  animation: ${fadein} 200ms ease-out 1;
```
jimmylee, line 23: nit: `fadeIn` and also update `const fadein` to `const fadeIn` above

components/core/FilePreviewBubble.js
```diff
@@ -0,0 +1,64 @@
...
+    <React.Fragment>
+      <div
+        css={STYLES_BUBBLE_ANCHOR}
+        onMouseEnter={() => setHover(true)}
+        onMouseLeave={() => setHover(false)}
+      >
+        {props.children}
+      </div>
+      {type && (
+        <div css={STYLES_FILE_PREVIEW_BUBBLE}>
```
jimmylee, line 56: Might want to leave a comment that the parent of this element needs to be `position: relative` or this bubble will fly off its position. or change the name of this component to `RelativeChildFilePreviewBubble`

components/core/FilePreviewBubble.js
```diff
@@ -0,0 +1,64 @@
...
+  padding: 8px;
+  border-radius: 4px;
+  background-color: ${Constants.system.black};
+  animation: ${fadein} 200ms ease-out 1;
+  width: 200px;
+  position: absolute;
+`;
+
+const STYLES_FILE_PREVIEW = css`
+  width: 100%;
```
jimmylee, line 29: `display: block`

components/core/FilePreviewBubble.js
```diff
@@ -0,0 +1,64 @@
...
+  height: 100%;
+`;
+
+const STYLES_BUBBLE_ANCHOR = css`
+  box-sizing: border-box;
```
jimmylee, line 34: `border-box` should be a natural style, you don't need to set it.

jimmylee (changes requested):

jimmylee:
Looking good! Just some details you should look at.

components/core/FilePreviewBubble.js
tarafanlin: what's the consequence for not writing `display: block` ?

jimmylee: block is width naturally 100%, height 100% doesn't do anything, and then the other two rules are not important anymore.

jimmylee:
LGTM

PR #378, opened 2020-10-21
@martinalong/new grid layout

martinalong:
- You can now edit your layouts freely and they won't get messed up when you add or delete from the slate!
- Items align to a dot grid for neatness
- Option to show titles
- New default layout style (will auto convert all old ones to this)
- Ability to repost things from other's slates to your own (will be lost if the original owner deletes their version)
- Ability to save a copy of anotehr person's file to your own files (won't be lost if the original owner deletes their version)

image


node_common/utilities.js
```diff
@@ -7,6 +7,7 @@ import JWT from "jsonwebtoken";
 import BCrypt from "bcrypt";
 
 import { Buckets, PrivateKey, Pow, Client, ThreadID } from "@textile/hub";
+import { CompressedPixelFormat } from "three";
```
jimmylee, line 10: What is this for?

pages/api/data/remove.js
```diff
@@ -3,33 +3,10 @@ import * as Utilities from "~/node_common/utilities";
...
-      w: 2,
-      h: 2,
-      minW: 2,
-      minH: 2,
-      // NOTE(jim): Library quirk thats required.
-      i: i.toString(),
-    };
-  });
-};
+import { read } from "fs";
```
jimmylee, line 6: What is this for?

server.js
```diff
@@ -117,9 +120,11 @@ app.prepare().then(async () => {
   server.all("/_/:a/:b", async (r, s) => handler(r, s, r.url));
 
   server.get("/:username", async (req, res) => {
+    let mobile = Validations.onMobile(req.headers["user-agent"]);
```
jimmylee, line 123: Probably should be using `~/common/window` for user agents, since Validations is for data integrity, but your call here.

PR #379, opened 2020-10-22
updating nav header and footer for external view

tarafanlin:
- adding CTA in the header for sign up/log in
- rearrange other actions to the footer
- apply max-width content and full-width background
Screen Shot 2020-10-21 at 9 34 32 PM

### DIFF
--- a/components/core/WebsitePrototypeHeader.js
+++ b/components/core/WebsitePrototypeHeader.js
@@ -71,8 +71,11 @@ const WebsitePrototypeHeader = (props) => {
           </a>
         </div>
         <div css={STYLES_RIGHT}>
+          <a css={STYLES_LINK} href="/_" style={{ marginRight: 24 }}>
+            Sign up
+          </a>
           <a css={STYLES_LINK} href="/_">
-            Sign up⼁Log in
+            Sign in
           </a>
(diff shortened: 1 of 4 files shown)


components/core/WebsitePrototypeHeader.js
```diff
@@ -1,97 +1,80 @@
...
+            css={STYLES_LINK}
+            href="/"
+            style={{ marginRight: 16, position: "relative", top: "1px" }}
+          >
+            <Logo style={{ height: 20 }} />
+          </a>
+        </div>
+        <div css={STYLES_RIGHT}>
+          <a css={STYLES_LINK} href="/_">
+            Sign up⼁Log in
```
jimmylee, line 75: Sign up,
Sign in

are our actions

jimmylee (changes requested):
Lets do a 24px gap for the bottom right items and see how that feels, if its too wide lets do 16px, same with the top we don't need the `|` seperator

tarafanlin:
@jimmylee new changes made
Screen Shot 2020-10-21 at 10 41 52 PM

jimmylee (approved):
LGTM

PR #380, opened 2020-10-22
minor fixes for the data scene

tarafanlin:
- remove "usage"
- fix file title overflow
Screen Shot 2020-10-21 at 9 58 21 PM

### DIFF
--- a/components/core/DataView.js
+++ b/components/core/DataView.js
@@ -67,6 +67,11 @@ const STYLES_LINK = css`
   overflow: hidden;
   text-overflow: ellipsis;
   white-space: nowrap;
+  max-width: 320px;
+
+  @media (max-width: ${Constants.sizes.tablet}px) {
+    max-width: 120px;
+  }
 `;
 
 const STYLES_VALUE = css`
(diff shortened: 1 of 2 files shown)


jimmylee:
Way to jump on this :) I'm not sure this fixes text-overflow though? Do you have a screenshot of the solution?

tarafanlin:
> Way to jump on this :) I'm not sure this fixes text-overflow though? Do you have a screenshot of the solution?

yep see the first file:
Screen Shot 2020-10-21 at 10 35 39 PM

jimmylee:
I suppose this solution works, Its better than what we have now, but its still better to use `flex-shrink: 0` anchor elements and then a min-width: 10% (for safari), width: 100%; and text-overflow: ellipsis

tarafanlin:
> I suppose this solution works, Its better than what we have now, but its still better to use `flex-shrink: 0` anchor elements and then a min-width: 10% (for safari), width: 100%; and text-overflow: ellipsis

ah ok, I can try this approach when I update the data table design

PR #385, opened 2020-10-26
new external slate view

tarafanlin:
- fix nav header and footer styling
- patch alert and header overlay
- adding counts
- more CTAs
- looks like this

Screen Shot 2020-10-25 at 10 08 26 PM


jimmylee (approved):
LGTM

PR #388, opened 2020-10-27
[WIP] external profile view

tarafanlin:
- new profile view info section
- will work on the slate preview blocks next step.
Screen Shot 2020-10-26 at 8 24 20 PM


common/constants.js
```diff
@@ -63,7 +63,8 @@ export const font = {
 };
 
 export const typescale = {
-  lvl0: `0.8rem`,
+  lvlN1: `12px`,
```
jimmylee, line 66: I think we can come up with better naming than this, Ideally I'd like to not downsize fonts so much but I can see the point. Why not try a value like `0.8rem` and use a converter here

http://www.standardista.com/px-to-rem-conversion-if-root-font-size-is-16px/

jimmylee (changes requested):

common/constants.js
tarafanlin: ok, made the convert, and lmk if you think we should increase the size a bit.

PR #389, opened 2020-10-27
@gndclouds/just tabs

gndclouds:
In an effort to make smaller PRs here is an update to the setting adding just tabs. PR #372 will now just focus an the autosave button. :-) 

image


jimmylee:
LGTM

PR #391, opened 2020-10-28
added search keybind

toastts:
adds keystroke listeners to appHeader on mount, search keybind right now is cmd/ctrl-f


martinalong:
This is awesome work @toastts !! I have no changes to suggest at all!

PR #394, opened 2020-10-29
new profile info section

tarafanlin:
new profile info section and some nit styling fixes for external slate & profile pages

Screen Shot 2020-10-29 at 12 27 13 PM

### DIFF
--- a/components/core/Profile.js
+++ b/components/core/Profile.js
@@ -16,12 +16,19 @@ const STYLES_PROFILE_INFO = css`
   display: flex;
   line-height: 1.3;
   margin: 0 auto;
+  width: 50%;
   @media (max-width: ${Constants.sizes.tablet}px) {
     width: 66.66%;
   }
   @media (max-width: ${Constants.sizes.mobile}px) {
     width: 100%;
   }
+`;
+
+const STYLES_PROFILE_INFO_INTERNAL = css`
+  display: flex;
+  line-height: 1.3;
+  margin: 0 auto;
 `;
 
 const STYLES_INFO = css`
@@ -61,7 +68,6 @@ const STYLES_NAME = css`
   margin-right: 24px;
   word-wrap: break-word;
   @media (max-width: ${Constants.sizes.mobile}px) {
-    font-size: ${Constants.typescale.lvl2};
     margin-bottom: 16px;
   }
 `;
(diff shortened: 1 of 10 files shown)


martinalong:
Do you think you can fix the alignment so that it also looks the same on the internal view? It looks centered here 
image

martinalong:
Also if you are on your own profile, it shouldn't show the follow button. I think the old follow button I had has that logic. You can look at that to see how it did that.

martinalong:
I'm guessing haris will also have some nits to make on alignment / spacing. So maybe send him a screenshot to get his opinion? Generally it's a good idea to do that when you make a change so that we can keep teh design consistent throughout Slate

jimmylee:
Love the initiative, would be nice to get @harisbutt involved with some of the review here.

tarafanlin:
added some fixes for the internal profile view here. preserving the buttons & functionality we have.
Screen Shot 2020-10-29 at 4 38 48 PM

martinalong:
Was adding this top bar here intentional? If not can you remove it from this view? Let me know if you need help doing that
image

components/core/Profile.js
```diff
@@ -8,40 +8,125 @@ import { ProcessedText } from "~/components/system/components/Typography";
...
   width: 100%;
-  margin-top: 16px;
+  padding: 24px 24px 0px 24px;
+`;
+
+const STYLES_PROFILE_INFO = css`
+  display: flex;
+  line-height: 1.3;
+  margin: 0 auto;
+  width: 50%;
```
martinalong, line 19: Can you make this:
``` 
width: 100%;
max-width: ${Constants.sizes.desktop}px;
```
That's the width and max width we have on the slate blocks that make up teh rest of the profile page

tarafanlin:
> Was adding this top bar here intentional? If not can you remove it from this view? Let me know if you need help doing that
> image

hmmm I missed that. will patch it

components/core/Profile.js
tarafanlin: cool! will do

martinalong:
And this is a weird edge case that happens on the external view.Not sure what's causing it
image

tarafanlin:
> And this is a weird edge case that happens on the external view when you have a really long one-word name.Not sure what's causing it to happen on external but not internal view
> image

hmmm interesting. I will take a look to see if I can reproduce this bug with long words

jimmylee:
You need `overflow-wrap: break-word;` @tarafanlin

tarafanlin:
> You need `overflow-wrap: break-word;` @tarafanlin

Went into a rabbit hole debugging this. 
What do you mean by "won't work if you have weird dom inline element nesting"? I tried `overflow-wrap` and it still has the issue.
Like @martinalong mentioned, I'm not sure why it works for the internal view but not an external view. Did I miss something? The styling I have is basically the same for those two in the inspector.
Screen Shot 2020-10-30 at 12 35 20 AM

Screen Shot 2020-10-30 at 12 33 37 AM

jimmylee:
Sometimes it depends on the parent element.

jimmylee:
I'll have a look tomorrow morning.

tarafanlin:
> I'll have a look tomorrow morning.

thanks cake!🙏

tarafanlin:
> Sometimes it depends on the parent element.

you're right it's the parent element. it's fixed!

martinalong:
Looks like it's fixed! 

I'm guessing haris will want you to align this follow button vertically with the name or move it to below the description like how you have it on the internal version. Might be good to do one last quick check with him 
image
Like this
image

martinalong:
And it looks like the white bar is still on the login page. Do you need help removing it?
image

tarafanlin:
> Looks like it's fixed!
> 
> I'm guessing haris will want you to align this follow button vertically with the name or move it to below the description like how you have it on the internal version. Might be good to do one last quick check with him
> image
> Like this
> image

ok will do!

tarafanlin:
> > I'm guessing haris will want you to align this follow button vertically with the name or move it to below the description like how you have it on the internal version. Might be good to do one last quick check with him

fixes 
Screen Shot 2020-10-31 at 12 57 29 PM
Screen Shot 2020-10-31 at 12 57 35 PM
Screen Shot 2020-10-31 at 12 59 21 PM

tarafanlin:
there is some issue with this branch so I created a new PR.

PR #398, opened 2020-10-31
Replaced rehydrate with webhooks

martinalong:
Removed rehydrate for everything except the initial login hydrate
Removed rehydrate button
Everything refreshes a lot faster now and we no longer need a lot of the loading spinners


jimmylee (approved):
LGTM, looks good, will test on main.

PR #413, opened 2020-11-09
adding default link preview images for files, slates and profiles

tarafanlin:


pages/_/profile.js
```diff
@@ -29,14 +32,14 @@ export default class ProfilePage extends React.Component {
     const title = this.props.creator ? `${this.props.creator.username}` : "404";
     const url = `https://slate.host/${title}`;
     const description = this.props.creator.data.body;
+    const image = this.props.creator.data.photo;
+
+    if (!image) {
```
jimmylee, line 37: if (Strings.isEmpty(image))

jimmylee (changes requested):
Love the initiative! Just one change requested.

pages/_/profile.js
tarafanlin: should I change the `!image` in slate.js to `(Strings.isEmpty(image))` as well?

Screen Shot 2020-11-09 at 3 41 26 PM

martinalong: yep!

martinalong:
image

For cases where you search through the objects and there are no objects, or no image objects (or all the image objects are too large to use as a preview image) it should show the default image that haris made as the preview image (both in the sidebar, and for the actual thing in slate.js)

tarafanlin:
> image
> 
> For cases where you search through the objects and there are no objects, or no image objects (or all the image objects are too large to use as a preview image) it should show the default image that haris made as the preview image (both in the sidebar, and for the actual thing in slate.js)

oh i see ok!

martinalong:
And also a thing I didn't think of but just realized: for cases where it's a private slate, it shouldn't show the preview section in the sidebar b/c there shouldn't be any preview if you link to the slate. So you coul dcheck if it's private, and if so hide that section entirely from the sidebar

martinalong:
> > image
> > For cases where you search through the objects and there are no objects, or no image objects (or all the image objects are too large to use as a preview image) it should show the default image that haris made as the preview image (both in the sidebar, and for the actual thing in slate.js)
> 
> oh i see ok!

image
here's an example of the case where there are files, but none of them are images

martinalong:
image
Dont' forget to set default_image for the sidebarsingleslatesettings too

tarafanlin:
done. the only problem is that our svg preview seems to break. so it doesn't really work for the case when svg is the only image in the document. should I track this down? or do you have a plan for this preview issue already?
Screen Shot 2020-11-09 at 3 59 09 PM
Screen Shot 2020-11-09 at 3 59 37 PM

martinalong:
image
one way you can make this code more efficient is: notice that lines 251-262 go through the list of items and check if any of them are valid preview images. But this is not necessary (and gets overwritten) if line 267 is true (if the link is for a specific item in a slate rather than the slate as a whole). So you can do this instead:
``` 
let image;
if(Strings.isEmpty(this.props.cid)) { //note this says "if strings.isempty", not "if !strings.isEmpty"
    image = preview; //in case there is a preview image set
    251-262 //handles the case in which there is no preveiw image set and you have to choose an appropriate one
    
} else {
    268-283 //handles the case where you're sharing a link to a specific item in a slate rather than the slate as a whole
}
263-265 //in case there's no appropriate preview image, set it to the default
```

tarafanlin:
> image
> one way you can make this code more efficient is: notice that lines 251-262 go through the list of items and check if any of them are valid preview images. But this is not necessary (and gets overwritten) if line 267 is true (if the link is for a specific item in a slate rather than the slate as a whole). So you can do this instead:
> 
> ```
> let image;
> if(Strings.isEmpty(this.props.cid)) { //note this says "if strings.isempty", not "if !strings.isEmpty"
>     image = preview; //in case there is a preview image set
>     251-262 //handles the case in which there is no preveiw image set and you have to choose an appropriate one
>     
> } else {
>     268-283 //handles the case where you're sharing a link to a specific item in a slate rather than the slate as a whole
> }
> 263-265 //in case there's no appropriate preview image, set it to the default
> ```

thanks! changes are made

martinalong:
Looks good! I'll merge this and then you can do some testing using the filecoin.onrender links. Be sure to check these cases:
- Slate with nothing in it (should show the default image)
- Slate with no image files in it (should show the default image)
- Slate with images files that are all over the 1mb size limit (should show the default image)
- Specific item in a slate (both image ones and non image ones which should default to the respective icon image)

tarafanlin:
> Looks good! I'll merge this and then you can do some testing using the filecoin.onrender links. Be sure to check these cases:
> 
> * Slate with nothing in it (should show the default image)
> * Slate with no image files in it (should show the default image)
> * Slate with images files that are all over the 1mb size limit (should show the default image)
> * Specific item in a slate (both image ones and non image ones which should default to the respective icon image)

yep will do!

PR #418, opened 2020-11-15
feat: add unity games

akuokojnr:
This PR detects if an upload is a Unity Game `.zip` file. If yes, it uses the zip extraction and upload endpoint on `Shovel` to handle its extraction and upload to Textile.

### DIFF
--- a/common/file-utilities.js
+++ b/common/file-utilities.js
@@ -3,6 +3,7 @@ import * as Store from "~/common/store";
 import * as Constants from "~/common/constants";
 import * as Credentials from "~/common/credentials";
 import * as Strings from "~/common/strings";
+import * as Validations from "~/common/validations";
 
 import { dispatchCustomEvent } from "~/common/custom-events";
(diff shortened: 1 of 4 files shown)


common/file-utilities.js
```diff
@@ -124,8 +129,9 @@ export const upload = async ({ file, context, bucketName, routes }) => {
   const storageDealRoute =
     routes && routes.storageDealUpload ? `${routes.storageDealUpload}/api/deal/` : null;
   const generalRoute = routes && routes.upload ? `${routes.upload}/api/data/` : null;
+  const zipUploadRoute = routes && routes.uploadZip ? `${routes.uploadZip}/api/data/zip/` : null;
 
-  if (!storageDealRoute || !generalRoute) {
```
jimmylee, line 128: We can probably clean this up later, but I don't want to block the work.

jimmylee (approved):
LGTM! Looks like a reasonable step 0. Will try this out now.

PR #432, opened 2020-11-20
Update: Moved UnityFrame from Canvas to Iframe

Aminejvm:
This PR fixes all the keyboard input related bugs

PR #433, opened 2020-11-20
Update: Moved UnityFrame from Canvas to Iframe #432

Aminejvm:
This PR fixes all the keyboard input related bugs

### DIFF
--- a/components/core/UnityFrame.js
+++ b/components/core/UnityFrame.js
@@ -7,48 +7,33 @@ const STYLES_CONTAINER = css`
   width: 100%;
 `;
 
-const loadScript = (url) =>
-  new Promise((res, rej) => {
-    // NOTE (Amine): Create script
-    const script = document.createElement("script");
-    script.src = url;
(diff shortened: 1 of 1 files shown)


jimmylee:
I am not sure I am happy with the move to the iFrame, great work @Aminejvm getting us to this point for discussion, @martinalong maybe we could explore for some slate objects what it looks like when they don't have the input fields? Since clicking will still work.

martinalong:
Yeah we can make it so the input fields are hidden for specific cases. Are each of these unity games of a the same file type? e.g. "application/pdf" or "application/unitygame". (or is there some other way to identify them?) If so we can check for that type and hide the sidebar with the inputs in those cases pretty easily.

jimmylee:
They will always be "application/unity" IIRC

Aminejvm:
Yes if we can hide inputs on this case3, we can definitely keep the current implementation, I just need to update the cleanup function to remove the unity script

PR #438, opened 2020-11-21
fix(deps): hub js hotfix

andrewxhill:
Signed-off-by: Andrew Hill <andrew@textile.io>


jimmylee (approved):
LGTM we'll find out if it just works :D

PR #447, opened 2020-11-23
temporary solution for username claiming

### LINKED ISSUE #436: Put username claiming in a more special section
Screen Shot 2020-11-20 at 11 12 50 AM

* Claiming a username is a big deal on Slate
* We don't want to give it the same UX treatment as a field you can just casually change.
* Can we put this somewhere else to give the field its importance?

tarafanlin:
- temporarily Resolves #436 – moving username claiming under the account tab, inform user the consequences of changing usernames
- - more specific error message for when the username is taken
Screen Shot 2020-11-22 at 9 05 07 PM
Screen Shot 2020-11-24 at 8 20 23 AM

Next step
- think the profile section in general could use some visual overhaul, I will discuss with Haris in our design chat for approaching this systematically with our design system

### DIFF
--- a/scenes/SceneEditAccount.js
+++ b/scenes/SceneEditAccount.js
@@ -167,14 +167,26 @@ export default class SceneEditAccount extends React.Component {
         },
       });
     } else if (response.error) {
-      dispatchCustomEvent({
-        name: "create-alert",
-        detail: {
-          alert: {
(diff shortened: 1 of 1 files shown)


jimmylee (approved):
LGTM, there are some things we can tweak later. But I think this deals with a problem I was concerned with.

PR #448, opened 2020-11-23
Fix: Cleanup Unity events after unmount

Aminejvm:
This PR fixes input issues even when the UnityFrame component unmounts, but the inputs still won't work while UnityFrame is in view.


jimmylee (approved):
LGTM, thanks :)

Issue #400, opened 2020-10-31
Implement shift + click


jimmylee:
# Purpose

<img width="1184" alt="Screen Shot 2020-10-31 at 3 15 53 PM" src="https://user-images.githubusercontent.com/310223/97791104-00ef8280-1b8c-11eb-95fb-2478af8d825e.png">

I should be able to shift + click and select many items at once to make it easier to manage my library.

# Deliverable

* Implement shift + click without causing any regressions.

akuokojnr:
@jimmylee we can close this issue now since it has been fixed with PR #456

PR #456, opened 2020-11-30
feat: range selection whilst pressing shift + click

akuokojnr:
This PR fixes Issue #400 

Now, the user can:

- select a range of items by pressing shift + click. This could be in either direction - up or down.

- can deselect a range of items by pressing shift + clicking on an item which has already been selected.

### DIFF
--- a/components/core/DataView.js
+++ b/components/core/DataView.js
@@ -749,7 +749,7 @@ export default class DataView extends React.Component {
           topRowStyle={{ padding: "0px 16px" }}
           onMouseEnter={(i) => this.setState({ hover: i })}
           onMouseLeave={() => this.setState({ hover: null })}
-          isShiftDown={this.state.isShiftDown}
+          isShiftDown={this.isShiftDown}
         />
         {footer}
         <input
--- a/components/core/DataView.js
+++ b/components/core/DataView.js
@@ -213,7 +213,10 @@ export default class DataView extends React.Component {
       window.addEventListener("resize", this.debounceInstance);
       window.addEventListener("keydown", this._handleKeyDown);
       window.addEventListener("keyup", this._handleKeyUp);
-      this.gridWrapperEl?.current.addEventListener("selectstart", this._handleSelectStart);
+
+      if (this.gridWrapperEl.current) {
+        this.gridWrapperEl.current.addEventListener("selectstart", this._handleSelectStart);
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


components/core/DataView.js
```diff
@@ -247,16 +263,95 @@ export default class DataView extends React.Component {
...
+  };
+
+  /* (NOTE):daniel This disable text selection while pressing shift key */
+  _handleSelectStart = (e) => {
+    if (this.state.isShiftDown) {
+      e.preventDefault();
+    }
+  };
+
+  _handleKeyUp = (e) => {
```
martinalong, line 304: It looks like sometimes it isn't properly registering key up and key down events. You can see this by putting these console.logs in and looking in your console as you press and release the shift key:
image

This is because you are checking `if(this.state.isShiftDown)`, but setting state is an asynchronous event (meaning it takes some time to finish). So if you key down then key up really quickly, it won't have enough time to set isShiftDown to true before it gets the next event to set it to false

I would recommend using a regular variable to store isShiftDown, rather than putting it in state. A regular variable won't have this issue (it sets instantly). You actually don't need to use a state variable here because isShiftDown does not affect how the page is rendered. The purpose of state variables is to ask React to "rerender the page every time this variable changes"

`this.isShiftDown = false` sets it as a regular variable as opposed to `this.setState({ isShiftDown: false })` which sets it as a part of state

components/core/DataView.js
```diff
@@ -247,16 +263,95 @@ export default class DataView extends React.Component {
...
+    }
+  };
+
+  _handleKeyDown = (e) => {
+    if (e.shiftKey && !this.state.isShiftDown) {
+      this.setState({ isShiftDown: true });
+    }
+  };
+
+  _handleGridCheckBox = (e, i) => {
```
martinalong, line 316: I don't think you need a separate _handleGridCheckBox and _handleTableCheckBox. These two do the same thing (though there's a small error in _handleTableCheckBox, so I'd keep just _handleGridCheckBox

components/core/DataView.js
```diff
@@ -247,16 +263,95 @@ export default class DataView extends React.Component {
...
+      });
+    }
+
     if (e.target.value === false) {
       delete checked[e.target.name];
-      this.setState({ checked });
+      this.setState({ checked, lastSelectedItemIndex: i });
       return;
     }
+
```
martinalong, line 283: This part from 278 - 287 seems redundant to the lines from 289 - 294

components/core/DataView.js
martinalong: You just have to pass in the index to it the same way that is done on line 597, since handleGridCheckBox uses a passed in value `i` rather than the name attached to the thing that fired the event

components/core/DataView.js
jimmylee: As we accumulate these hotkeys we should also start using this https://github.com/ianstormtaylor/is-hotkey

akuokojnr: @jimmylee is-hotkey library works with a combination of two or more keys. In this case, we are only checking if the shift key has been pressed. Hence, it will always return false in this case.

martinalong:
Looks great!

PR #460, opened 2020-12-01
patch broken link for explore slates

tarafanlin:
the link for exploreSlates were broken, now it's fixed.

### DIFF
--- a/server.js
+++ b/server.js
@@ -255,6 +255,11 @@ app.prepare().then(async () => {
           "0ba92c73-92e7-4b00-900e-afae4856c9ea",
         ],
       });
+
+      for (let exploreSlate of exploreSlates) {
+        let user = await Data.getUserById({ id: exploreSlate.data.ownerId });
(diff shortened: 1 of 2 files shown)


server.js
```diff
@@ -255,6 +255,11 @@ app.prepare().then(async () => {
           "0ba92c73-92e7-4b00-900e-afae4856c9ea",
         ],
       });
+
+      for (let exploreSlate of exploreSlates) {
+        let user = await Data.getUserById({ id: exploreSlate.data.ownerId });
+        exploreSlate.username = user.username;
```
jimmylee, line 261: In the future its better to keep some sense of structure like

`slate.user.username`
`slate.owner.username`

so you can always assume that if there is an attached user, you can grab the username. This concept can be addressed by introducing something like "serialization"

you can address this later, but just want to point it out now

jimmylee (approved):
LGTM

PR #463, opened 2020-12-03
hide actions when action bar is working

### LINKED ISSUE #442: Don't show other actions when the action bar is working.
Screen Shot 2020-11-21 at 2 36 39 PM

* We should hide other options since the action bar is already doing work.

tarafanlin:
- hide actions when action bar is working Resolve #442 
- add loader indicator in slate layout

in dataview:
Screen Shot 2020-12-02 at 9 32 57 PM

in slatelayout:
Screen Shot 2020-12-02 at 9 26 10 PM

### DIFF
--- a/components/core/CarouselSidebarSlate.js
+++ b/components/core/CarouselSidebarSlate.js
@@ -164,9 +164,11 @@ export default class CarouselSidebarSlate extends React.Component {
     showFile: true,
     unsavedChanges: false,
     pickerLoading: false,
+    loading: false,
   };
 
   componentDidMount = () => {
+    window.addEventListener("slate-global-carousel-loading", this._handleSetLoading);
     if (this.props.isOwner && !this.props.external) {
       this.debounceInstance = this.debounce(() => this._handleSave(), 3000);
       let isPublic = false;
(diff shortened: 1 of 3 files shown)


martinalong:
Hmm when I tested it out, the action bar just disappears as soon as I click the buttons on the action bar rather than staying there and seeing the loading spinner. 

Also for the action bar on slate layout, can you make it so it is just one loading spinner that you see instead of two side by side?

tarafanlin:
> Hmm when I tested it out, the action bar just disappears as soon as I click the buttons on the action bar rather than staying there and seeing the loading spinner.
> 
> Also for the action bar on slate layout, can you make it so it is just one loading spinner that you see instead of two side by side?

To the first one, I think this is how it works previously. I didn't make any changes. But I can make it so that it stays there.

And yeah I can make the loading spinner work just for "delete files". I'm assuming "remove" works instantaneously so there is no need to make a loader?

martinalong:
When is the loader spinner in the action bar visible then? Like when you click delete directly from the slate object preview?

Yeah seeing it now I think it would be nice if it stayed there and had the loading spinner as it was deleting. That way you know it's in progress rather than not getting any update for a long time

tarafanlin:
hide actions on deleted objects and hide checkboxes for all objects when deleting
Screen Shot 2020-12-03 at 8 11 18 PM
Screen Shot 2020-12-03 at 9 04 27 PM

martinalong:
The action bar looks good! I did notice that on the slate page, the loader spinner looked like it was below the file while the action was being completed. Is it possible to make it like the data view, where the spinner is where the checkbox was previously?
image
image

tarafanlin:
> The action bar looks good! I did notice that on the slate page, the loader spinner looked like it was below the file while the action was being completed. Is it possible to make it like the data view, where the spinner is where the checkbox was previously?
> image
> image

I will get back to you hopefully tonight on this!

tarafanlin:
the loader appears at the bottom right corner now!  @martinalong 
Screen Shot 2020-12-07 at 4 24 04 PM

PR #464, opened 2020-12-03
Prismjs ssr component

akuokojnr:
This PR resolves issue #404.  

It removes prismjs global styles and refactors `CodeBlock` component to handle all use cases of prism.js


components/system/CodeBlock.js
```diff
@@ -27,6 +24,98 @@ const STYLES_CODE_BLOCK = css`
       height: 0;
     }
   }
+
+  :not(pre) > code,
```
jimmylee, line 28: So I think a key thing with this next iteration of the CodeBlock is to inline all of the styles for prism instead of using  global styles or scoped styles. So I'm going to have to block this PR on that (Sorry it took me so long to get to it).

jimmylee (changes requested):
So have a look at my comment, basically I wouldn't mind if you found a vendor library that inlines the styles instead of applying them globally or scoped.

components/system/CodeBlock.js
akuokojnr: Sure. Not a problem. I guess I did not understand clearly was the demands were.

PR #468, opened 2020-12-08
fix: get unity game config and loader from data prop 

akuokojnr:
This PR ensures that irrespective of the folder structure of the unity it is still able to load on Slate.

It passes the two things essential to load the game, i.e. `unityGameConfig` and `unityGameLoader`, as props to the `UnityFrame.js` component.

Thanks to @tarafanlin for spotting this issue.

### DIFF
--- a/components/core/SlateMediaObject.js
+++ b/components/core/SlateMediaObject.js
@@ -59,7 +59,9 @@ export default class SlateMediaObject extends React.Component {
     const url = this.props.data.url;
     const type = this.props.data.type ? this.props.data.type : "LEGACY_NO_TYPE";
     const playType = typeMap[type] ? typeMap[type] : type;
-
+    const unityGameConfig = this.props.data.unityGameConfig;
+    const unityGameLoader = this.props.data.unityGameLoader;
+    console.log(this.props.data);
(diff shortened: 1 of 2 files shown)


components/core/SlateMediaObject.js
```diff
@@ -94,7 +96,9 @@ export default class SlateMediaObject extends React.Component {
 
     // TODO(jim): We will need to revisit this later.
     if (type.startsWith("application/unity")) {
-      return <UnityFrame url={url} />;
+      return (
```
jimmylee, line 99: Can you put these values in the if guard code? Since nothing else needs it.

components/core/UnityFrame.js
```diff
@@ -26,17 +26,17 @@ const _cleanScripts = () => {
   });
 };
 
-const UnityFrame = ({ url }) => {
+const UnityFrame = ({ url, unityGameLoader, unityGameConfig }) => {
```
jimmylee, line 29: Awesome. this is great.

jimmylee (approved):
LGTM, can you follow up on the nit? thanks :)

components/core/SlateMediaObject.js
akuokojnr: Sure

PR #470, opened 2020-12-08
Activity feed

martinalong:
image


jimmylee:
Activity feed looks a lot like a Slate lol

node_common/monitor.js
```diff
@@ -57,7 +55,7 @@ const createSlateActivityForEachSubscriber = async ({ userId, data }) => {
       Data.createActivity({
         userId: s.owner_user_id,
         data: {
-          type: "ANOTHER_USER_CREATED_SLATE",
+          type: "OTHER_USER_CREATE_SLATE",
```
jimmylee, line 58: might need to come up with a universal term as useful as `viewer` at somepoint. like `observer`

PR #471, opened 2020-12-09
nits styling

tarafanlin:
fix the styling for index header & external previewblocks
Screen Shot 2020-12-08 at 10 21 06 PM
Screen Shot 2020-12-08 at 10 21 30 PM


jimmylee (approved):
Why are we taking out the ProcessedText?

tarafanlin:
> Why are we taking out the ProcessedText?

`text-overflow: ellipsis` doesn't work with ProccessedText for some reason.
And I think for preview blocks, we don't really need to apply links to the body text. The whole block is already a link.

PR #479, opened 2020-12-17
edit file name

tarafanlin:
allow people to change the name of data objects: click on the data object name to edit

Screen Shot 2020-12-17 at 2 44 26 PM

Screen Shot 2020-12-17 at 2 44 31 PM

### DIFF
--- a/components/core/CarouselSidebarData.js
+++ b/components/core/CarouselSidebarData.js
@@ -189,21 +189,21 @@ const STYLES_AUTOSAVE = css`
   justify-content: baseline;
   color: ${Constants.system.yellow};
   opacity: 0;
-  margin: 8px 0 32px 0;
+  margin: 26px 24px;
 
   @keyframes autosave {
     0% {
       height: 0;
       opacity: 0;
-      margin-left: 12px;
+      margin-left: 0;
     }
     10% {
       opacity: 1;
-      margin-left: 0;
+      margin-left: 24px;
     }
     90% {
       opacity: 1;
-      margin-left: 0;
+      margin-left: 24px;
     }
     100% {
       opacity: 0;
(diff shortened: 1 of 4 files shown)


martinalong:
Works perfectly! Just a few UI suggestions:

- Maybe make it more clear that clicking on it will make it editable (maybe cursor: pointer? or something else as well?)
- If you edit it then click outside (which turns it off edit mode), it should update and save immediately rather than waiting a few seconds till the debounce
- Also, rather than allowing people to type capitals and then lowercasing upon save, you should just convert things to lowercase as they type. Example:
```
  _handleUsernameChange = (e) => {
    this.setState({ [e.target.name]: e.target.value.toLowerCase() });
  };
```
- (Not essential, but a nice to have) It's ideal if we don't have the UI shifting up and down when the "autosaved" appears then disappears. So maybe put it in this spot to the right of the "231.14 KB"
image
- (Also just a nice to have) maybe can save and exit the input field upon hitting enter as well

components/core/CarouselSidebarData.js
```diff
@@ -338,11 +418,37 @@ export default class CarouselSidebarData extends React.Component {
...
+              />
+            </Boundary>
+          ) : (
+            <span
+              css={STYLES_META_TITLE}
+              target="_blank"
+              // href={Strings.getCIDGatewayURL(cid)}
+              onClick={this._handleEditFilename}
+            >
+              {name}
```
martinalong, line 439: I'm still not seeing the filename change reflected immediately after editing. You can fix that by changing this line to {this.state.name} rather than name. That way this will always be the same value as the input field

martinalong:
The other thing I forgot to say in the last review: 

Up to you whether you want to do this. The main thing it would affect is cases where you upload it to a slate and we show the file ending in that slate like here where I changed the filename to Tuna (getting rid of the .jpg) and so it no longer shows .JPG in the slate
image

If you want to preserve the file extension (the .jpg part of filename.jpg) and not allow people to edit that, the way to do that is: when it goes into editing mode, you "separate" the filename into two parts: filename.jpg -> filename + jpg and you make only the first part editable, and then recombine the extension with the new filename afterwards

components/core/CarouselSidebarData.js
tarafanlin: yeah I removed the debounce but there is still a bit lagging. thanks for pointing this out!

martinalong:
LGTM! And good thinking on using the .file instead of .name for extensions! Saved us a lot of time haha

PR #480, opened 2020-12-17
fix: game root link

akuokojnr:


jimmylee (approved):
I think this is right

PR #482, opened 2020-12-20
@tarafanlin/cleanup

tarafanlin:
- swap out project wor feature slate
- code cleanup
- autosave ui nits

Screen Shot 2020-12-20 at 12 49 52 PM

### DIFF
--- a/components/core/CarouselSidebarData.js
+++ b/components/core/CarouselSidebarData.js
@@ -191,26 +191,24 @@ const STYLES_AUTOSAVE = css`
   opacity: 0;
   margin: 26px 24px;
 
-  @keyframes autosave {
+  @keyframes slate-animations-autosave {
     0% {
-      height: 0;
       opacity: 0;
-      margin-left: 0;
+      transform: translateX(0);
     }
     10% {
       opacity: 1;
-      margin-left: 24px;
+      transform: translateX(12px);
     }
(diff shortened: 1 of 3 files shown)


components/core/CarouselSidebarData.js
```diff
@@ -191,26 +191,24 @@ const STYLES_AUTOSAVE = css`
...
-      margin-left: 24px;
+      transform: translateX(12px);
     }
     100% {
       opacity: 0;
-      height: 0;
     }
   }
-  animation: autosave 4000ms ease;
+  animation: slate-animations-autosave 4000ms ease;
```
jimmylee, line 211: This would be a nice cleanup task to consolidate all of these at some point, really nice though.

components/core/CarouselSidebarSlate.js
```diff
@@ -92,24 +92,24 @@ const STYLES_AUTOSAVE = css`
   position: absolute;
   opacity: 0;
 
-  @keyframes autosave {
+  @keyframes slate-animations-autosave {
```
jimmylee, line 95: Whats awesome about these animations are they become global, so we can add them like we add other global styles in the future. Again, a really awesome cleanup task for some later time.

jimmylee (approved):
Love it. thanks!

PR #492, opened 2021-01-18
new profile

tarafanlin:
new profile minus the profile edit drawer. 
I will make the profile edit feature in a separate PR because this is already a big one.

**THANKS @martinalong FOR ALL THE NOTES AND HELPS WITH DEBUGGING. They're very helpful.**

Screen Shot 2021-01-18 at 9 36 34 AM

### DIFF
--- a/components/core/Profile.js
+++ b/components/core/Profile.js
@@ -1,13 +1,18 @@
 import * as React from "react";
 import * as Constants from "~/common/constants";
 import * as Strings from "~/common/strings";
+import * as SVG from "~/common/svg";
 
 import { css } from "@emotion/react";
 
 import ProcessedText from "~/components/core/ProcessedText";
 import SlatePreviewBlocks from "~/components/core/SlatePreviewBlock";
 import CTATransition from "~/components/core/CTATransition";
+import DataView from "~/components/core/DataView";
+import EmptyState from "~/components/core/EmptyState";
+
 import { SceneUtils } from "three";
+import { TabGroup, SecondaryTabGroup } from "~/components/core/TabGroup";
 
 const STYLES_PROFILE_BACKGROUND = css`
   background-color: ${Constants.system.white};
@@ -20,7 +25,7 @@ const STYLES_PROFILE_BACKGROUND = css`
 
 const STYLES_PROFILE_INTERNAL = css`
   width: 100%;
-  padding: 64px 56px 0px 56px;
+  padding: 0px 56px 0px 56px;
   overflow-wrap: break-word;
   white-space: pre-wrap;
   @media (max-width: ${Constants.sizes.mobile}px) {
@@ -134,12 +139,18 @@ const STYLES_BUTTON = css`
 export default class Profile extends React.Component {
   state = {
     exploreSlates: [],
+    tab: 1,
+    view: 0,
+    slateTab: 0,
+    peerTab: 0,
   };
 
   render() {
     const external = !this.props.onAction;
     let data = this.props.creator ? this.props.creator : this.props.data;
     let exploreSlates = this.props.exploreSlates;
+
+    console.log(this.props);
 
     let total = 0;
     for (let slate of data.slates) {
@@ -178,7 +189,6 @@ export default class Profile extends React.Component {
             </div>
           </div>
         </div>
-
         {this.state.visible && (
           <div>
             <CTATransition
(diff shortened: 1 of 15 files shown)


jimmylee:
YAY NICE

components/system/components/GlobalCarousel.js
```diff
@@ -11,6 +11,7 @@ import { Alert } from "~/components/core/Alert";
 import CarouselSidebarSlate from "~/components/core/CarouselSidebarSlate";
 import CarouselSidebarData from "~/components/core/CarouselSidebarData";
 import SlateMediaObject from "~/components/core/SlateMediaObject";
+import { CompressedPixelFormat } from "three";
```
jimmylee, line 14: New line

components/system/components/GlobalCarousel.js
```diff
@@ -124,6 +125,12 @@ export class GlobalCarousel extends React.Component {
     window.removeEventListener("slate-global-close-carousel", this._handleClose);
   };
 
+  componentDidUpdate = (prevProps) => {
+    if (this.state.index >= (this.props.objects?.length || 0)) {
```
jimmylee, line 129: Nice use of the language feature! Still feel like this condition is a little brittle, can we improve it?

components/system/components/GlobalCarousel.js
```diff
@@ -155,20 +162,16 @@ export class GlobalCarousel extends React.Component {
   };
 
   _handleOpen = (e) => {
-    let carouselType =
-      !this.props.current ||
-      (this.props.current &&
-        (this.props.current.decorator === "FOLDER" || this.props.current.decorator === "ACTIVITY"))
-        ? "data"
-        : "slate";
+    if (e.detail.index < 0 || e.detail.index >= this.props.objects.length) {
```
jimmylee, line 165: This condition, albeit useful, is a little difficult to read, can we improve it?

scenes/SceneProfile.js
```diff
@@ -98,62 +91,15 @@ export default class SceneProfile extends React.Component {
...
-            Unfollow
-          </ButtonSecondary>
-        ) : (
-          <ButtonPrimary style={{ marginRight: 8 }} onClick={this._handleFollow}>
-            Follow
-          </ButtonPrimary>
-        )}
-      </div>
-    );
+    //FOR TARA: this is the change that'll allow it to update "automatically" if you're viewing your own profile and make changes
```
jimmylee, line 94: Please stick to our comment convention

`NOTE(Tara):`

jimmylee (approved):
LGTM, I mostly approve, please have @martinalong review.

I have a greater issue outside of this PR, where I don't think any of the `string` states should be camel case or lower case, instead I think they should be `ALL_CAPS_LIKE_I_AM_YELLING_AT_YOU` for a number of reasons

* I don't want to confuse english strings in the codebase with potential types.
* As we move towards `type` safety in the codebase (maybe we do a TS conversion in the near future), it will be easier to seek and change all of the states.
* Optically its more clear.

Its not a huge priority, but as you're working on the codebase I'd like to see this change 🙏

scenes/SceneProfile.js
martinalong: @jimmylee these were meant to be temporary comments I left for tara so she knows what I changed and why. Meant for them to be easily found by CTRL + F and deleted after so I used a different format

martinalong: @tarafanlin think you missed this one. I'll take it out. This one is important to know since this was the cause of the stale data issues you were having. Was just a small change that made all the difference

tarafanlin: oh yes sorry I only CTRL+F inside of profile.js, forgot about this one. I will review the comment and take it out!

components/core/Profile.js
```diff
@@ -103,78 +101,495 @@ const STYLES_DESCRIPTION = css`
...
+            onClick={() => {
+              this.props.onAction({
+                type: "NAVIGATE",
+                value: this.props.sceneId,
+                scene: "PROFILE",
+                data: relation.user,
+              });
+            }}
+            external={this.props.external}
+            url={`https://slate.host/${relation.user.username}`}
```
martinalong, line 440: I'll make this change, but in the future, it's better to do links as url={`/${relation.user.username}`} so that they work on dev server as well. Otherwise when you test out these links on dev, it'll redirect you to slate.host rather than localhost

Also simpler in case the domain ever changes in the future

tarafanlin: I see ok

components/core/Profile.js
```diff
@@ -103,78 +101,495 @@ const STYLES_DESCRIPTION = css`
...
+            css={STYLES_DIRECTORY_PROFILE_IMAGE}
+            style={{ backgroundImage: `url(${user.data.photo})` }}
+          />
+          <span css={STYLES_DIRECTORY_NAME}>
+            {user.data.name || `@${user.username}`}
+            {message ? <span css={STYLES_MESSAGE}>{message}</span> : null}
+          </span>
+        </div>
+      )}
+      {button}
```
martinalong, line 231: This menu should be hidden on external, so "{external ? null : button}"

components/core/Profile.js
```diff
@@ -103,78 +101,495 @@ const STYLES_DESCRIPTION = css`
...
+      : !!this.props.viewer.subscriptions.filter((entry) => {
+          return entry.target_user_id === this.props.data.id;
+        }).length,
+  };
+
+  componentDidMount = async () => {
+    await this.filterByVisibility();
+  };
+
+  componentDidUpdate = async () => {
```
martinalong, line 262: Currently, if you go from one profile to another, it'll stay on the same tab b/c react perceives it as the same component (doesn't restart state or call component did mount). So to manually take the user to the default slates tab, I'm adding this
```
  componentDidUpdate = async (prevProps) => {
    if (this.props.creator?.id !== prevProps.creator?.id) {
      this.setState({ tab: 1, view: 0, fileTab: 0, slateTab: 0, peerTab: 0 });
    }
```

tarafanlin: hmmm, good catch! not sure if prevProps work tho. remember the bug we talked about?

components/core/Profile.js
tarafanlin: why should we hide the menu on external?

martinalong: Because the menu has two things in it: "follow/unfollow" and "copyURL". Copy url is unnecessary and I'll take it out of all menus since we now have nice urls that you can copy directly from the url bar. Follow /unfollow is not alllowed for external

components/core/Profile.js
martinalong: i tested it out and it works fine for this case

PR #496, opened 2021-01-24
add api endpoints for social and explore & tab routing

tarafanlin:
**NOTES - something we could discuss further to see if there is a better approach** 
- for fetching exploreSlates, @martinalong and I decided to move it to the top of server.js, so that it will only fetch once for all users until the next server reboot
- tab routing is set up for internal but not external. everything still works tho. @martinalong might walk me through query param later
- when you navigate to the peers tab / following slates tab, there would be about 1 sec delay, for fetching social stuff
Screen Shot 2021-01-23 at 5 28 00 PM


common/actions.js
```diff
@@ -98,7 +98,15 @@ export const getSlateById = async (data) => {
 
 export const getSlatesByIds = async (data) => {
   await Websockets.checkWebsocket();
-  return await returnJSON(`/api/slates/get`, {
+  return await returnJSON(`/api/v1/explore`, {
```
jimmylee, line 101: Are you sure nothing else is using `getSlatesByIds`?

jimmylee: Its okay to change the internals of a function if nothing else is using it. But then this function is named incorrectly then?

common/actions.js
```diff
@@ -98,7 +98,15 @@ export const getSlateById = async (data) => {
...
+    ...DEFAULT_OPTIONS,
+    body: JSON.stringify({ data }),
+  });
+};
+
+export const getSocial = async (data) => {
+  await Websockets.checkWebsocket();
```
jimmylee, line 108: Why do we need to checkWebsocket here? (cc: @martinalong)

components/core/Application.js
```diff
@@ -75,7 +75,9 @@ const SCENES = {
   ACTIVITY: <SceneActivity tab={0} />,
   EXPLORE: <SceneActivity tab={1} />,
   DIRECTORY: <SceneDirectory />,
-  PROFILE: <SceneProfile />,
+  PROFILE_FILES: <SceneProfile tab={0} />,
```
jimmylee, line 78: Nice!

pages/_/profile.js
```diff
@@ -51,7 +52,14 @@ export default class ProfilePage extends React.Component {
       <WebsitePrototypeWrapper title={title} description={description} url={url} image={image}>
         <WebsitePrototypeHeader />
         <div css={STYLES_ROOT}>
-          <Profile {...this.props} buttons={buttons} isOwner={false} external />
+          <Profile
```
jimmylee, line 55: Nice!

pages/api/slates/get.js
```diff
@@ -26,27 +26,14 @@ export default async (req, res) => {
...
-  } else {
-    const response = await Data.getSlateById({ id: req.body.data.id });
-    if (!response) {
-      return res.status(404).send({ decorator: "SERVER_GET_SLATE_NOT_FOUND", error: true });
-    }
-
-    if (response.error) {
-      return res.status(500).send({ decorator: "SERVER_GET_SLATE_NOT_FOUND", error: true });
-    }
+  const response = await Data.getSlateById({ id: req.body.data.id });
```
jimmylee, line 29: This is confusing, this old endpoint seems to have nothing wrong with it?

pages/api/v1/social.js
```diff
@@ -0,0 +1,48 @@
...
+
+  const r2 = await Serializers.doSubscribers({
+    users: [],
+    slates: [],
+    subscribers,
+    serializedUsersMap: r1.serializedUsersMap,
+    serializedSlatesMap: r1.serializedSlatesMap,
+  });
+
+  if (!subscriptions) {
```
jimmylee, line 27: Would be better to cancel the request sooner if the first one failed., the decorator looks like its copy and pasted and should be specific to this call

pages/api/slates/get.js
jimmylee: I don't think you need to modify this endpoint at all. You should just keep it the way it is.

jimmylee (changes requested):
Looking good! Nice work, I would not change `api/slates/get` and just add new functions instead. It seems like `api/slates/get` does exactly what it's supposed to

In fact if it is a publicly exposed route, you could call it from the client instead of making a whole new explore endpoint (if explore doesn't do anything special)

Does that make sense? Let me know if you need clarification.

common/actions.js
tarafanlin: actually I'm just following the format here. I could remove it.
but when do we need to checkWebsocket then?

jimmylee: We use WebSockets to deliver real time updates back to the browser really fast. We check if the WebSocket is alive so when we perform this action, we know we will receive a socket update instead of having to manually update the state of the react application.

I'm assuming these WebSocket checks ensure that we're still connected/reconnect, you might not need it for social!

pages/api/slates/get.js
tarafanlin: hmmm so original I set up this for explore. 
But I thought it might be better to separate it out bc it errors on line 7 when passing in an array.

tarafanlin: let me double check with Martina on this one

common/actions.js
martinalong: Yeah exactly what jim said. You need it if the endpoint it calls has something like "hydratePartialSlates" or "hydratePartialLibrary". That means it uses websockets to update the UI, and you want to make sure it checks (and if needed, reconnects) websockets before it makes the call. In this case, not necessary since the info just comes back in the response

common/actions.js
martinalong: So currently nothing is using this endpoint, but generally it's not a good idea to changes things like this. I don't think you are using this either tara so maybe change it back so that its behavior is predictable. 

FYI when you call Data.getSlatesById, it's not the same as this (which would be called with Actions.getSlatesById). Data.getSlatesById calls node_common/data/methods/get-slates-by-ids. Data is on the backend, Actions is on the front end

common/actions.js
martinalong: If you take a look at the other calls, you'll see that some of the functions in action.js have checkWebsocket and some don't. I left a note at the top explaining it

common/actions.js
```diff
@@ -98,7 +98,15 @@ export const getSlateById = async (data) => {
...
+  return await returnJSON(`/api/v1/social`, {
```
martinalong, line 109: I think since this is a user related api call, it should go under api/users, not api/v1 (v1 is for external apis I believe, like the ones the chrome extension uses). And maybe name it get-peers or something?

pages/api/slates/get.js
martinalong: Yeah don't change this. I think other things use the array version so this will break it. Basically I made this endpoint so that it works for both get (single) and get (multiple). Can you show me how it's erroring?

pages/api/v1/explore.js
```diff
@@ -0,0 +1,14 @@
+import * as Data from "~/node_common/data";
```
martinalong, line 1: since nothing is using this endpoint this can you remove this?

pages/api/v1/social.js
martinalong: @tarafanlin so when you're working with the backend code, you want to send different decorators for each error so that you know which error is happening when you get an error back. Take a look at the other files to get an idea of what I mean. 

The other thing is whenever you add a new decorator, you also should go to common/messages.js and add a corresponding message. This is how know which error message to show the user: when the frontend receives an error back from an api call, it passes it through this Events.hasError(response) function which triggers an alert if there is an error. That alert looks up the decorator in messages.js, and shows the corresponding alert message (or a generic alert message if not found)

server.js
```diff
@@ -61,6 +61,48 @@ const EXTERNAL_RESOURCES = {
...
+        "0ba92c73-92e7-4b00-900e-afae4856c9ea",
+      ],
+    });
+
+    for (let exploreSlate of exploreSlates) {
+      let user = await Data.getUserById({ id: exploreSlate.data.ownerId });
+      exploreSlate.username = user.username;
+    }
+  } else {
+    exploreSlates = await Data.getSlatesByIds({
```
martinalong, line 85: Keep in mind that this part may error for other people. Other people don't necessarily have the same slates and slateIds as you in their database. If you wanted to be safe, I would comment this out (so that you can use it to test if needed still) and return an empty array if not in production

common/actions.js
tarafanlin: I see. for the name maybe get-social? it's fetching peers and following slates

common/actions.js
tarafanlin: yep I know, I was just thinking that we could probably keep the endpoint just in case. 
I will change it back

pages/api/v1/social.js
tarafanlin: oh I finally see what you guys mean. 
the english word "subscriptions" and "subscribers" look too similar to me, I didn't realize the two are using the same decorator.

common/messages.js
```diff
@@ -188,6 +188,10 @@ export const error = {
   GET_USER_BY_USERNAME:
     "We weren't able to fetch information on that user. Please check your input",
   USER_NOT_FOUND: "We aren't able to locate that user at the moment. Please try again",
+  SERVER_USER_SUBSCRIPTIONS_NOT_FOUND:
+    "We weren't able to fetch information on that user. Please try again later",
```
jimmylee, line 192: Nice! Thanks for doing this :)

pages/api/users/get-social.js
```diff
@@ -0,0 +1,49 @@
...
+  }
+
+  if (subscribers.error) {
+    return res.status(500).send({ decorator: "SERVER_USER_SUBSCRIBERS_NOT_FOUND", error: true });
+  }
+
+  let serializedUsersMap = { [req.body.data.userId]: req.body.data };
+  let serializedSlatesMap = {};
+
+  const r1 = await Serializers.doSubscriptions({
```
jimmylee, line 28: I think this will work! Nice way to lay it all out :-)

pages/api/users/get-social.js
```diff
@@ -0,0 +1,49 @@
...
+    users: [],
+    slates: [],
+    subscriptions,
+    serializedUsersMap,
+    serializedSlatesMap,
+  });
+
+  const r2 = await Serializers.doSubscribers({
```
jimmylee, line 36: In the future when people are copy & pasting some of my code, it might be better to be more redundant, this reads like a sentence:

`const serializedSubscribers = await Serializers.doSubscribers();`

This is my fault, `r2` is a really bad variable name. But for the future, let's move towards naming things where people can read the behavior like they were reading a text message (within reason).

server.js
```diff
@@ -61,6 +61,49 @@ const EXTERNAL_RESOURCES = {
...
+  //   });
+
+  //   for (let exploreSlate of exploreSlates) {
+  //     let user = await Data.getUserById({ id: exploreSlate.data.ownerId });
+  //     exploreSlate.username = user.username;
+  //   }
+  // }
+};
+
+fetchExploreSlates();
```
jimmylee, line 105: This is a clever idea! We'll see how it goes :-)

jimmylee: why not `exploreSlates = await fetchExploreSlates()`; and `let exploreSlates = []`

jimmylee (changes requested):
Just one change I think you should make which is writing a function that doesn't return undefined and instead returns the data for assignment. The current function assigns a memory address and probably works but it does it as a side effect which isn't very resilient code.

server.js
tarafanlin: hmm I think "await" is only allowed in async function. 
if we put `exploreSlates = await fetchExploreSlates()` on top, it will error right?
unless you mean otherwise?

jimmylee: You should do it here then: 

```
  const listenServer = server.listen(Environment.PORT, async (e) => {
    if (e) throw e;
    Websocket.create();

    NodeLogging.log(`started on http://localhost:${Environment.PORT}`);
    
    exploreSlates = await fetchExploreSlates();
  });
  ```

jimmylee (approved):
LGTM, go ahead and merge @tarafanlin looks safe

PR #498, opened 2021-01-27
tab routing

tarafanlin:


martinalong:
Summary: Tara added routing by window.history.pushState to profiles, so that you can click back/forward within profiles but the URL will stay the same

PR #501, opened 2021-02-02
fix: unity game zip download ui improvements

akuokojnr:
This PR makes a UI improvement to the game download. Since a network request is made to Textile to get the game files and zip them, we need to show that the download is in progress (one which may take a few seconds to complete).

To do this, the download icon changes to a loading spinner and the "download" changes to "downloading" until it is completed.


jimmylee (approved):
Awesome PR. Nice one ✨

PR #503, opened 2021-02-04
feat: show user online status

akuokojnr:
First iteration on user presence for multiplayer slate. Currently, it shows if a user is online. See screenshot below.
active-profile

Corresponding PR for fiji here: https://github.com/slate-engineering/fiji/pull/1

TO-DO

- [x] Sync with @harisbutt and @tarafanlin about designs for presence (online/offline) for an active user and other users that the current user follows in the "Following" tab. Currently, I am simply showing a green border for when a user is online.

### DIFF
--- a/common/browser-websockets.js
+++ b/common/browser-websockets.js
@@ -63,7 +63,6 @@ export const init = ({ resource = "", viewer, onUpdate, onNewActiveUser }) => {
       const response = JSON.parse(event.data);
       type = response.type;
       data = response.data;
-      console.log(data);
     } catch (e) {
       console.log(e);
     }
--- a/common/browser-websockets.js
+++ b/common/browser-websockets.js
@@ -8,7 +8,7 @@ let savedResource = null;
 let savedViewer = null;
 let savedOnUpdate = null;
 
-export const init = ({ resource = "", viewer, onUpdate }) => {
+export const init = ({ resource = "", viewer, onUpdate, handleActiveUsers }) => {
   savedResource = resource;
   savedViewer = viewer;
   savedOnUpdate = onUpdate;
@@ -63,6 +63,7 @@ export const init = ({ resource = "", viewer, onUpdate }) => {
       const response = JSON.parse(event.data);
       type = response.type;
       data = response.data;
+      console.log(data);
     } catch (e) {
       console.log(e);
     }
@@ -77,6 +78,10 @@ export const init = ({ resource = "", viewer, onUpdate }) => {
 
     if (type === "UPDATE") {
       onUpdate(data);
+    }
+
+    if (type === "UPDATE_USERS_ONLINE") {
+      handleActiveUsers(data);
     }
   });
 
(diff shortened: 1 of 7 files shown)


components/core/Profile.js
```diff
@@ -508,7 +523,10 @@ export default class Profile extends React.Component {
           <div css={STYLES_PROFILE_INFO}>
             <div
               css={STYLES_PROFILE_IMAGE}
-              style={{ backgroundImage: `url('${creator.data.photo}')` }}
+              style={{
+                backgroundImage: `url('${creator.data.photo}')`,
+                borderColor: this.state.isOnline ? `green` : `white`,
```
jimmylee, line 528: Lets make sure we use constants values for this, I know its tedious but keeps it sane.

components/core/Profile.js
```diff
@@ -352,7 +355,19 @@ export default class Profile extends React.Component {
     });
   };
 
+  checkStatus = () => {
+    const activeUsers = this.props.activeUsers;
+    const userId = this.props.data.id;
+
+    if (activeUsers && activeUsers.includes(userId)) {
```
jimmylee, line 362:
```js
this.setState({ isOnline: activeUsers && activeUsers.includes(userId) });
```

common/browser-websockets.js
```diff
@@ -8,7 +8,7 @@ let savedResource = null;
 let savedViewer = null;
 let savedOnUpdate = null;
 
-export const init = ({ resource = "", viewer, onUpdate }) => {
+export const init = ({ resource = "", viewer, onUpdate, handleActiveUsers }) => {
```
jimmylee, line 11: `onNewActiveUser`?

jimmylee (changes requested):

jimmylee:
Nice work @akuokojnr, love seeing progress.

akuokojnr:
Thank you, Jim! Preciate it!

jimmylee (approved):
LGTM, we should only have one green so we can reconcile that later or something :-)

components/core/Application.js
```diff
@@ -209,6 +210,7 @@ export default class ApplicationPage extends React.Component {
         resource: this.props.resources.pubsub,
         viewer: this.state.viewer,
         onUpdate: this._handleUpdateViewer,
+        onNewActiveUser: this._handleActiveUsers,
```
jimmylee, line 213: `onNewActiveUser: this._handleNewActiveUser` would be the convention I like. You can change this later.

akuokojnr: Sure. I'll change it now.

akuokojnr:
> LGTM, we should only have one green so we can reconcile that later or something :-)

I used that because I think that we're going to have to think about how to display user online status on the the profile. Since this is just for the first iteration, I did not want to add a new color

akuokojnr:
Offline state
offline

Online state
online

jimmylee:
LGTM. this is awesome. thanks.

PR #508, opened 2021-02-05
Add PDF previews using react pdf

Xuefeng-Zhu:


jimmylee:
Hey @Xuefeng-Zhu this is awesome! Thanks! I think I made a mistake earlier...

So I think I should clarify some expectations...

## The Task

I'd like you to get a design system component going as a `page`, you should be able to get this all running locally.

The way you do it as follows:

1) Create a page! Here is an example of a file you can copy and paste and rename. https://github.com/filecoin-project/slate/blob/main/pages/_/system/buttons.js
2) Add the page to the list: Here is the component the list lives in https://github.com/filecoin-project/slate/blob/main/components/system/SystemPage.js#L173
3) Now verify! You should be able to see a new page generated for whatever you named your page as.
4) On the page, show us a method that works for mobile and iPad for rendering PDFs from IPFS. You should use this PDF I sourced for you: https://slate.textile.io/ipfs/bafybeibyeaznm4iaungi6cfxpy75mnkpn64c3zw2kgq7qkpvecvqlrpw6u

Please try to keep the file structure the same as other files, I will help advise you along the way.

Hope this helps clarify some things. Thanks again, means a lot that you would come to help us!

Xuefeng-Zhu:
test on desktop and mobile
image

image

jimmylee:
Ah this is perfect! Thank you @Xuefeng-Zhu

jimmylee (approved):
LGTM

PR #511, opened 2021-02-09
basic link drag and drop

whyrusleeping:
I dont really know how to run slate locally to test this, but something like this worked locally in isolation, so maybe we can get this across the finish line and get past my bad javascripting.


jimmylee (approved):
LGTM, I will update the naming on this thing but works.

PR #518, opened 2021-02-10
feat: use new pdf renderer

akuokojnr:
What better way to render PDF in slate than to use the rendering library created by Adobe, creators of the PDF?

This PR replaces react-pdf with Adobe Document Cloud PDF Embed API (https://www.adobe.io/apis/documentcloud/dcsdk/pdf-embed.html). Initial tests reveal that this is very robust, does not crash and is very responsive across devices and browsers.

 #### TODO

- [x] Improve styling in global carousel
- [ ] Make changes to systems page

### DIFF
--- a/components/system/components/PDFViewer.js
+++ b/components/system/components/PDFViewer.js
@@ -14,18 +14,17 @@ const STYLES_DOCUMENT = css`
   overflow-y: scroll;
   -ms-overflow-style: none;
   scrollbar-width: none;
-
-  ::-webkit-scrollbar {
-    display: none;
-  }
+  display: grid;
+  place-items: center;
 `;
 
-const STYLES_PAGE = css`
-  margin-bottom: 20px;
+const STYLES_WRAPPER = css`
+  width: calc(100% - 120px);
+  max-width: 850px;
 `;
 
-const PDFViewer = ({ file, clientId, ...rest }) => {
-  const viewerConfig = {
+const PDFViewer = ({ file, clientId, viewerConfig, ...rest }) => {
+  viewerConfig = {
     showAnnotationTools: false,
     enableFormFilling: false,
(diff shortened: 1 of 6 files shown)


akuokojnr:
#### Desktop
desktop

#### Mobile
iphone-x

akuokojnr:
@jimmylee In this implementation, we need an API key to render the PDF file. Due to this, I am wondering if we should still add this component to our design system and share with other. I would love to know how you think that I should approach this.

akuokojnr:
Scrollbar is added this time around because of issue #522

components/system/components/PDFViewer.js
```diff
@@ -17,41 +14,63 @@ const STYLES_DOCUMENT = css`
...
+
+  React.useEffect(() => {
+    document.addEventListener("adobe_dc_view_sdk.ready", initializeViewer);
+  });
+
+  return (
+    <>
+      <Head>
+        <script src="https://documentcloud.adobe.com/view-sdk/main.js" />
+      </Head>
```
cwaring, line 65: Worth double checking if this will this insert the script every time this component is rendered/mounted. I would typically recommend using a HOC (such as [react-async-script](https://github.com/dozoisch/react-async-script)) to fetch any external dependencies so that the lifecycle is managed better.

cwaring:
Great that this works better than `react-pdf` however we should also consider the privacy (and dependency) cost of adding an external library directly from adobe into an app like Slate. Perhaps @jimmylee has a stronger stance on this?

components/system/components/PDFViewer.js
akuokojnr: Okay. I'm gonna look into this

akuokojnr:
@cwaring Do you have alternatives in mind?

PR #525, opened 2021-02-11
get user display name and profile photo for slate chrome

tarafanlin:


martinalong:
Hey! Looks good. The one thing I'd recomend changing is using the standard format for user that we use across slate. Otherwise it'll get confusing fast which place uses what data format. Aka pass in "user" instead of "name" and "profilePhoto" (also because not every user has a name. A lot only have username and never added a name)

And "user" can take a format like this:

```
user: {
    username: xxx,
    data: {
         name: xxx,
         body: xxx,
         photo: xxx,
    }
}
```

tarafanlin:
> Hey! Looks good. The one thing I'd recomend changing is using the standard format for user that we use across slate. Otherwise it'll get confusing fast which place uses what data format. Aka pass in "user" instead of "name" and "profilePhoto" (also because not every user has a name. A lot only have username and never added a name)
> 
> And "user" can take a format like this:
> 
> ```
> user: {
>     username: xxx,
>     data: {
>          name: xxx,
>          body: xxx,
>          photo: xxx,
>     }
> }
> ```

fixed!

Issue #517, opened 2021-02-10
Add shared linting config and git hook automation for Slate projects


cwaring:
### Scope
As our team/project grows and more contributions roll in we should invest a little time to configure the base CI automation for code and style linting, this will collectively make it quicker and easier to review PRs, catch minor bugs and also help avoid the topic of style by guiding contributors automagically.

### Possible solutions
I've used a similar combo in other projects to great success. Once configured it is really a set and forget experience.
- [husky](https://github.com/typicode/husky) for custom git repo hooks, pre-commit/post-merge checks 
- [lint-staged](https://github.com/okonet/lint-staged) to operate on the current head of edits for a branch
- [eslint](https://eslint.org/) with a [react-app](https://www.npmjs.com/package/eslint-config-react-app) and/or [standardjs](https://standardjs.com/) config
- [stylelint](https://stylelint.io/) for css and css-in-js blocks
- [prettier](https://prettier.io/) with eslint config to tie everything together

cwaring:
This is integrated without stylelint for the meantime, calling it complete

PR #526, opened 2021-02-12
chore: add eslint config and husky lint-staged checks

cwaring:
Opening for further discussion 💁‍♂️

This PR adds base `eslint` config along with auto prettier formatting. Currently reporting **1170** problems mostly from `react/prop-types` because there are no type definitions across the app.

There are a few a11y things that we should fix and many of the rest are style related that I can fix automatically. If you'd like to see the report simply run `npm run lint` from the root of the app.

I can configure it to suppress any rules that we deem overkill for now, such as types until we come to refactor time however I'm mostly in favour of all the other recommendations. Contribs, let me know your preference!

[report.txt](https://github.com/filecoin-project/slate/files/5972006/report.txt)

ref: #517

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -1,0 +1,3 @@
+node_modules
+dist
+vendor
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -21,6 +21,7 @@ module.exports = {
   extends: [
     "eslint:recommended",
     "plugin:react/recommended",
+    "plugin:react-hooks/recommended",
     "plugin:jsx-a11y/recommended",
     "plugin:prettier/recommended", // last item for prettier config to take priority
   ],
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -14,7 +14,6 @@ module.exports = {
   },
   env: {
     browser: true,
-    amd: true,
     node: true,
     es2021: true,
   },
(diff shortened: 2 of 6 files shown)


jimmylee:
Lets not worry about types for a while and make sure the team adheres to

## Case 1

Always de-structure JavaScript objects when they are used as function arguments

```
const someFunction = ({ a, b, c, d }) => { console.log(a); }
```

## Case 2

Always pass in arguments as new javascript objects

```
functionDoThing({ a: source.a, b: other.b, c: true });
```

When we do make the move to types, we should probably just use TypeScript and call it a day, I don't think we should use flow and we should only use types in parts of the codebase where we won't be dropping `any` everywhere or in > 60% of the cases.

cwaring:
@jimmylee nice, agree. I think both of those cases will have to be managed by hand at PR time because it's a valid scenario to call a function with multiple args and perhaps a little more robust because we can set defaults at the method level. Maybe we can write a plugin later if we want more control.

For now I've setup the option to `prefer-destructing`, which will trigger against assigned objects/arrays and disabled checks for `react/prop-types`, agree typescript it totally best at this stage but I don't want to give everyone too much of a hard time.

This is ready to merge now and will only throw warnings/errors for files part of new commits and husky 🐶 will take care of any fixes it can automatically, also if the npm deps have changed `npm install` will run after merge ⚡

cwaring:
Merging for now, we can continue to tweak as we further to define our style 🚀

PR #529, opened 2021-02-12
feat: drag to download

Aminejvm:
Drag file to desktop to download it
Currently only works on chrome (it's the only browser that supports this behavior)


jimmylee (approved):
LGTM, seems okay!

PR #531, opened 2021-02-15
revert pdf preview changes

akuokojnr:
This PR reverts PDF preview changes made in Slate. Now, we use `object` element to render PDF files on desktop but open a PDF file in a new tab on mobile devices. This behaviour will automatically trigger download on mobile devices per my tests.


jimmylee (approved):
LGTM

PR #532, opened 2021-02-15
fix: blank carousel on mobile

akuokojnr:
Since we open pdf in a new tab on mobile, the carousel remains blank. This PR fixes it by automatically closing the carousel after opening the pdf in a new tab on mobile.

### DIFF
--- a/components/core/SlateMediaObject.js
+++ b/components/core/SlateMediaObject.js
@@ -1,6 +1,7 @@
 import * as React from "react";
 import * as Constants from "~/common/constants";
 import * as Validations from "~/common/validations";
+import * as Events from "~/common/custom-events";
 
 import UnityFrame from "~/components/core/UnityFrame";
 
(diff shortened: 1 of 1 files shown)


components/core/SlateMediaObject.js
```diff
@@ -1,6 +1,7 @@
 import * as React from "react";
 import * as Constants from "~/common/constants";
 import * as Validations from "~/common/validations";
+import * as Events from "~/common/custom-events";
```
jimmylee, line 4: What is `Events` for I don't see you using it

jimmylee (changes requested):

jimmylee (approved):
LGTM I'm wrong. sorry.

components/core/SlateMediaObject.js
jimmylee: My bad I saw it. Oops.

PR #534, opened 2021-02-16
fix: clicking on Unity doesn't close the carrousel anymore

Aminejvm:


components/core/UnityFrame.js
```diff
@@ -46,7 +46,11 @@ const UnityFrame = ({ url, unityGameLoader, unityGameConfig }) => {
     };
   }, []);
 
-  return <div id="unityContainer" css={STYLES_CONTAINER} />;
+  return (
+    <div css={STYLES_CONTAINER} onClick={(e) => e.stopPropagation()}>
```
jimmylee, line 50: I think this is okay for now, there is a better way I'll teach some other time involving a traversal up the tree and some domain knowledge around bubbling and capture.

jimmylee (approved):

Issue #535, opened 2021-02-16
add a changelog and semver release cycle to Slate


cwaring:
## Purpose
It would be great to provide a `CHANGELOG` to showcase updates and features landing in Slate, we could also include this as part of the interface to aid discovery of new Slate features and educate others how to build web3 apps.

## Deliverable
My suggestion is to automate this by applying a [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/) schema to our commits and [standard-version
](https://github.com/conventional-changelog/standard-version) to generate the changelog doc. This would require squashing PRs into the `main` branch for things to work smoothly and keep history clean.

Also open to other ideas, feel free to leave info below.

┆Issue is synchronized with this [Notion page](https://www.notion.so/add-a-changelog-and-semver-release-cycle-to-Slate-713efc76d20b437a85ffff2c85741553) by [Unito](https://www.unito.io)

PR #537, opened 2021-02-16
chore: update deps and add package-lock

cwaring:
As prep for versioned releases we should have a package locked bundle to make it easier to isolate changes and create dependable snapshots.

ref: #536 & #535


cwaring:
does anybody know why we're using a forked release of [`react-blurhash`](https://github.com/filecoin-project/slate/blob/main/package.json#L53)? can I switch it to the latest npm published release?

jimmylee (approved):
LGTM, yeah, makes sense!

PR #538, opened 2021-02-16
feat: markdown file reader

Aminejvm:
markdown

Currently supports the following element ( headings, paragraph, image, link, lists )


components/core/MarkdownFrame.js
```diff
@@ -0,0 +1,105 @@
+import * as React from "react";
+import { css } from "@emotion/react";
```
jimmylee, line 2: Please reorder imports to match other files.

components/core/MarkdownFrame.js
```diff
@@ -0,0 +1,105 @@
...
+  width: 100%;
+  height: 100%;
+  overflow-y: scroll;
+  will-change: transform;
+  color: ${Constants.system.white};
+`;
+
+const STYLES_BODY = css`
+  width: 100%;
+  max-width: 687px;
```
jimmylee, line 21: Where does `687px` come from?

jimmylee (approved):
LGTM, just some questions and things you might want to polish and be more intentional about. Slate seems to be designed on a pt4 system so make sure measurements are multiples of 4 or 8 or 12

components/core/MarkdownFrame.js
Aminejvm: I used it as a way to ensure we have a maximum of 70 characters in one line

components/core/MarkdownFrame.js
Aminejvm: I'll make sure to do that

Issue #488, opened 2021-01-04
add shorturl option to image view


autonome:
# Purpose

http gateway urls already give up decentralization, so we should also provide short urls for easy sharing.

# Deliverable

* image view on slate.host now has a cid, gateway url and a short url option

* api response for objects in a slate should include short url

# Dependencies

N/A

cwaring:
I'd love to discuss the motivation of this further. Personally I think it's a great to keep the CID in the url, this makes the object hash very accessible, aids with awareness of a CID and also provides a great way to browse manually if I already have a CID of an object that I want to find on Slate.

I know we could add some of that functionality back in other ways but it's a strong stance to give it pride of place in URL. I see a future where I could access a Slate object directly via `slate://{CID}` and the UI will show me the object and all the connections with boards and users 💭

jimmylee:
Are we saying that we can make `slate://` happen? If so, that would be insanely cool. I can't wait for

`slate://username:slatename:cid`

how insane would that be

cwaring:
Yeah, we can register applinks via a desktop/electron wrapper or iOS/Android (with a bit of effort) to enable direct deep links. 

It's probably still better to only expose http links and redirect for pure accessibility but just [as an example](https://blog.branch.io/deep-linking-from-url-and-uri-schemes-to-universal-links-app-links-and-beyond/) for how something like this would work and why thinking carefully about our URL structure early on is important 😁

autonome:
To be clear: I'm not saying Slate should change any existing URLs. I'm asking for a short URL option to be added in user share contexts, like the properties pane or any share icon.

There's no future world where winning is people looking at CIDs all the time. Visible CIDs may play some transitional role as a visual affordance in some contexts, but the endgame is most certainly not us staring at them.

We live in an HTTP world, and will be in a transitional stage for a lonnnnng time. Given the positioning of Slate as a product using the lingua franca as an adoption strategy, getting more people using Slate > getting people to use Slate to look at CIDs. So a short URL for places where shorter URLs will encourage more clicking is good, eg text messages and other chat platforms.

[...]

PR #539, opened 2021-02-16
feat: add short urls for slate objects

akuokojnr:
This PR fixes issue #488. It replaces long cids with short url.

TO DO

- [ ] Test all actions (download, add to slate, remove from slate, delete, etc) to ensure that they have not been affected
- [ ] Redirect all cid urls to new short urls
- [ ] Add `shortid(s)`  to slate objects which have been created already

### DIFF
--- a/components/core/Application.js
+++ b/components/core/Application.js
@@ -507,10 +507,10 @@ export default class ApplicationPage extends React.Component {
     const id = Window.getQueryParameterByName("scene");
     const user = Window.getQueryParameterByName("user");
     const slate = Window.getQueryParameterByName("slate");
-    const cid = Window.getQueryParameterByName("cid");
+    const shortid = Window.getQueryParameterByName("shortid");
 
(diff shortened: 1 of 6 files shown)


akuokojnr:
Preview:
short-url

cwaring:
Hi @akuokojnr, I've added a few thoughts in original issue. Keen to see an expansion on this before axing CIDs from the URLs, unless I'm mistaken about this issue completely 😅

akuokojnr:
Thanks for weighing in on this @cwaring I have been thinking about this today actually. I think that we can have short urls without axing `cids` from the current url. In my opinion, this is the way to go.

I think that we can add the short url to the slate data object and make it available when needed. This will mean that there should be a dedicated server route (e.g. `slate.host/{shortid}`) as well.

akuokojnr:
I have read your comments on the original issue and I agree with you. They're valid points.

PR #541, opened 2021-02-16
privacy and filetype filters for files folder

toastts:
pretty good working version for now, but im gonna change it to be general file types (Audio Video etc.) soon since that seems to be more user-friendly

### DIFF
--- a/scenes/SceneFilesFolder.js
+++ b/scenes/SceneFilesFolder.js
@@ -42,19 +42,20 @@ const STYLES_TOOLTIP_ANCHOR = css`
   z-index: ${Constants.zindex.tooltip};
 `;
 
-//toast: setting both of these to 50% width might be janky but it works
 const STYLES_FILETYPE_TOOLTIP = css`
   display: flex;
   flex-direction: column;
   justify-content: space-around;
-  height: 250px;
+  height: 260px;
 `;
 
 const STYLES_PRIVACY_TOOLTIP = css`
   display: flex;
   flex-direction: column;
   justify-content: space-around;
-  height: 110px;
+  height: 115px;
+  font-family: ${Constants.font.medium};
+  font-size: 16px;
 `;
 
 export default class SceneFilesFolder extends React.Component {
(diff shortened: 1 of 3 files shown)


scenes/SceneFilesFolder.js
```diff
@@ -15,9 +21,138 @@ import EmptyState from "~/components/core/EmptyState";
...
+const STYLES_FILTERS_CONTAINER = css`
+  height: 60px;
+  display: flex;
+  flex-direction: row;
+  justify-content: flex-end;
+  align-items: flex-start;
+`;
+
+const STYLES_TOOLTIP_ANCHOR = css`
+  border: 1px solid #f2f2f2;
```
jimmylee, line 33: Leave yourself a TODO here so you can add that to constants when we unify this into some design system

scenes/SceneFilesFolder.js
```diff
@@ -15,9 +21,138 @@ import EmptyState from "~/components/core/EmptyState";
...
+
 export default class SceneFilesFolder extends React.Component {
   state = {
     view: 0,
+    filetypeTooltip: false,
+    privacyTooltip: false,
+    filters: {},
+    filtersActive: false,
+    privacy: "All",
+    files: this.props.viewer.library[0].children,
```
jimmylee, line 69: This will scream if the viewer isn't here, so check if viewer exists first.

scenes/SceneFilesFolder.js
```diff
@@ -15,9 +21,138 @@ import EmptyState from "~/components/core/EmptyState";
...
+
+  _handlePrivacyTooltip = () => {
+    this.setState({ privacyTooltip: !this.state.privacyTooltip });
+  };
+
+  _handlePrivacyFilter = (filter) => {
+    const viewer = this.props.viewer;
+    const filtered = getPublicAndPrivateFiles({ viewer });
+    if (filter === "All") {
+      this.setState({ privacy: "All" }, this._filterFiles);
```
jimmylee, line 84: Represent these states in all lowercase or uppercase to save yourself the one by one grief you could have in the future.

jimmylee (approved):
LGTM, nice one, this is a great PR. you did good @toastts

scenes/SceneFilesFolder.js
toastts: they're sentence-cased for the button so I can just put {this.state.privacy} without manipulating the text, it has to be done ;-;

PR #550, opened 2021-02-19
fix: broken file upload to slate on slate page

akuokojnr:
When you open a particular slate page, it is expected that when you upload a file, it should be added to that slate. Currently, it does not. This PR fixes that.


jimmylee (approved):
Interesting, I'll have to look if this causes any side effects but seems okay. LGTM

PR #552, opened 2021-02-19
fix: broken global carousel on page refresh or when user visit slate object url

akuokojnr:
When the carousel is open and you refresh the page, it does not open again. It just opens the slate page containing the slate object. Also, when you copy the address to a slate object and paste in a new tab, the carousel does not open. This PR fixes these related issues.

### DIFF
--- a/scenes/SceneSlate.js
+++ b/scenes/SceneSlate.js
@@ -313,16 +313,16 @@ class SlatePage extends React.Component {
 
     /* NOTE(daniel): If user was redirected to this page, the cid of the slate object will exist in the page props. 
     We'll use the cid to open the global carousel */
-    if (cid) {
-      const index = this.getItemIndexByCID(cid);
-
-      Events.dispatchCustomEvent({
-        name: "slate-global-open-carousel",
-        detail: { index },
-      });
-
-      return;
+    if (Strings.isEmpty(cid)) {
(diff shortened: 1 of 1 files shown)


scenes/SceneSlate.js
```diff
@@ -302,6 +306,33 @@ class SlatePage extends React.Component {
...
   };
 
+  _handleURLRedirect = () => {
+    const {
+      page: { cid },
+    } = this.props;
+
+    /* NOTE(daniel): If user was redirected to this page, the cid of the slate object will exist in the page props. 
+    We'll use the cid to open the global carousel */
+    if (cid) {
```
jimmylee, line 316: Can you change this to

`if (Strings.isEmpty(cid)) { return null }`

And then the bottom code doesn't have to be in a `{}` if closure.

scenes/SceneSlate.js
```diff
@@ -302,6 +306,33 @@ class SlatePage extends React.Component {
...
+      const index = this.getItemIndexByCID(cid);
```
jimmylee, line 317: Can you catch the case if the `cid` doesn't exist here?

jimmylee (changes requested):

jimmylee (approved):

PR #554, opened 2021-02-19
fix slate block title cut off

tarafanlin:
tiny PR fixing the text cutoff here
Screen Shot 2021-02-19 at 9 30 28 AM

should work now: 
Screen Shot 2021-02-19 at 9 33 40 AM


jimmylee (approved):
LGTM

Issue #521, opened 2021-02-10
feature: unique object urls & history push/pop overlay urls to the location bar


cwaring:
![2021-02-10 at 15 32 49](https://user-images.githubusercontent.com/106938/107531989-512ba700-6bb5-11eb-9edb-783df71647da.gif)

# Purpose

Currently on the activity view (and perhaps elsewhere) when you view a file in the overlay it is not possible to copy the location bar providing a route back to that item, this makes it hard to share a direct link to a slate object (outside of the slate board view).

# Deliverable

Add a routing pattern so that all public objects have a unique link that is suitable for sharing, preferably complete with og meta previews so they are pretty on social media.

This should also help provide an additional entry path for acquiring new users as we can present an opportunity to sign up to save the object to their own slate board.

martinalong:
Yeah so we have og links for the file once you click into it and view it in the slate, but haven't added it for when you open carousel in activity. Can easily add it though since the info should be all there. Someone else can take this on or I can do it once I get back to web in a few weeks

martinalong:
Awesome job @akuokojnr !

PR #559, opened 2021-02-22
feat: show slate object url in location bar on activity page

akuokojnr:
This PR fixes issue #521. We will now show the slate object url in the location bar on activity page when the user opens the object. This will make sharing slate easier by copying the slate object from the location bar.


jimmylee (approved):
LGTM! Nice one Daniel :-)

PR #575, opened 2021-02-24
feat: url capture and mediaObject frames

cwaring:
# UI Logic
webloc-logic

For a given link: https://protocol.ai/

Capture the following normalised data: 

```JSON
{
  "title": "Protocol Labs",
  "description": "Protocol Labs is building the next generation of the internet.",
  "lang": "en",
  "author": null,
  "publisher": "Protocol Labs",
  "image": {
    "url": "https://protocol.ai/images/og-default.jpg",
    "type": "jpg",
    "size": 136032,
    "height": 630,
    "width": 1200,
    "size_pretty": "136 kB"
  },
  "url": "https://protocol.ai",
  "logo": {
    "url": "https://logo.clearbit.com/protocol.ai",
    "type": "png",
    "size": 8773,
    "height": 128,
    "width": 128,
    "size_pretty": "8.77 kB"
  }
}
```

## Status: WIP
Frontend:

* [x] create default frame plug-in examples
* [x] capture thumbnail image from og data
* [x] design default media object style
* [ ] x-browser/os testing
* [ ] docs/description

Stretch:

* [x] capture foundation NFT type
* [ ] capture foundation NFT author
* [x] design foundation NFT layout

Backend:

* [x] Setup auth proxy stream for mq
* [x] Only capture url to IPFS file object
* [ ] subscribe to mq pro tier

ref: #545

PR #577, opened 2021-02-25
fix: carousel editor & ui fixes

cwaring:
Small ui/logic fixes discussed on Slack. (cherry-picked from #575)

### DIFF
--- a/components/core/CarouselSidebarData.js
+++ b/components/core/CarouselSidebarData.js
@@ -226,7 +226,7 @@ const STYLES_SPINNER = css`
   height: 24px;
 `;
 
-export const FileTypeDefaultPreview = () => {
+export const FileTypeDefaultPreview = (props) => {
   if (props.type && props.type.startsWith("video/")) {
     return DEFAULT_VIDEO;
   }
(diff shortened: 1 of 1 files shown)


components/core/CarouselSidebarData.js
```diff
@@ -472,20 +474,20 @@ class CarouselSidebarData extends React.Component {
       <div key="s-2" style={{ marginBottom: 80 }}>
         <div css={STYLES_META}>
           {this.state.isEditing ? (
-            <Boundary enabled onOutsideRectEvent={this._handleEditFilename}>
+            <Boundary enabled onOutsideRectEvent={this._handleEditFilename.bind(this)}>
```
jimmylee, line 477: You don't need to `.bind(this)` if you're using an arrow function.

components/core/CarouselSidebarData.js
```diff
@@ -472,20 +474,20 @@ class CarouselSidebarData extends React.Component {
...
               <Input
                 full
                 value={this.state.name}
                 name="name"
-                onChange={this._handleChange}
+                onChange={this._handleChange.bind(this)}
```
jimmylee, line 482: Same here `You don't need to `.bind(this)` if you're using an arrow function.`

components/core/CarouselSidebarData.js
```diff
@@ -412,7 +413,8 @@ class CarouselSidebarData extends React.Component {
     }
   };
 
-  _handleEditFilename = async () => {
+  _handleEditFilename = () => {
```
jimmylee, line 416: No bind is necessary.

jimmylee (changes requested):

cwaring:
Super strange, there was a case yesterday where `this` was pointing to a closure function and unable to access the class properties which was causing these bugs, I suspected it was because these elements were being pushed into an array before rendering. Now I've restarted my env and I'm unable to recreate it 🤷‍♂️

I'll close this for now and file it under babel weirdness.

PR #600, opened 2021-03-02
feat: multi-file download

### LINKED ISSUE #588: Multi file download
# Purpose

- Right now you can only download one file at a time
- Would like to be able to select multiple files then download them
- And add a button to download a whole slate

# Deliverable

- Will need to zip the files first then download it

Aminejvm:
This PR resolves #588 
[Progress so far on notion](https://www.notion.so/slatesystem/b3ea5ea2f0124a098651546b8f6ba40e?v=e21432223f0f48cb9478dae30109749e&p=c94a6c4e387947829932752ff8901415)


common/user-behaviors.js
```diff
@@ -334,6 +334,62 @@ export const downloadZip = async (file) => {
   }
 };
 
+const _nativeDownload = ({ url, onError }) => {
+  const iframe = document.createElement("iframe");
+  iframe.style.display = "none";
+  iframe.src = url;
+
+  const ERROR_MESSAGE = "SLATE_DOWNLOAD_ERROR";
+  const hanldeIframeErrors = (e) => {
```
martinalong, line 343: seems to be a misspelling here of "handle"

martinalong: nice solve for the error detection though!

Aminejvm: you're right @martinalong! I'll fix it

martinalong: don't worry though, i'll make the fix and just merge it

PR #605, opened 2021-03-03
improvement: Unity parallel download

Aminejvm:
This PR just improved the download speed of Unity games

### DIFF
--- a/common/actions.js
+++ b/common/actions.js
@@ -347,7 +347,7 @@ export const getActivity = async (data) => {
 };
 
 export const getZipFilePaths = async (data) => {
-  return await returnJSON(`../api/zip/get-paths`, {
+  return await returnJSON(`/api/zip/get-paths`, {
     ...DEFAULT_OPTIONS,
     body: JSON.stringify({ data }),
   });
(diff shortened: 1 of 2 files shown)


common/actions.js
```diff
@@ -347,7 +347,7 @@ export const getActivity = async (data) => {
 };
 
 export const getZipFilePaths = async (data) => {
-  return await returnJSON(`api/zip/get-paths`, {
+  return await returnJSON(`../api/zip/get-paths`, {
```
jimmylee, line 350: How is this possible?

jimmylee (changes requested):
I'm not sure how you're able to get a path like that...? That can't possibly work? Unless I misunderstand something.

common/actions.js
Aminejvm: https://denmchenry.com/posts/relative-urls/, actually the browser interprets it as the parent URL, I think why it didn't work previously is because it didn't have `/` before `api`, I thought it did have the `/` that's why I added the double dots. I'll update it right now

common/actions.js
```diff
@@ -347,7 +347,7 @@ export const getActivity = async (data) => {
 };
 
 export const getZipFilePaths = async (data) => {
-  return await returnJSON(`api/zip/get-paths`, {
+  return await returnJSON(`/api/zip/get-paths`, {
```
jimmylee, line 350: This is correct. much better.

jimmylee (approved):

Issue #586, opened 2021-02-28
Add tagging to slates and files


martinalong:
# Purpose

- Add tagging to slates and files(e.g.: "pets", "food", "design", "architecture")
- Adding tagging means that users can auto-categorize slates and files for us, allowing activity to be more categorizeable as well

# Deliverable

- Sync with @harisbutt on how the design of this should look
- Add a place in the slate settings and file carousel to add tags
- Add a place to display the tags under the description of the slate and files
- Add an entry in the database in the data blob for a slate that has its tags in it
- Then @martinalong will add tag searchability afterwards

# Dependencies

 - #582 let's wait for this first, since this will make this task much simpler (since you don't have to keep files synced between data and slates)

PR #615, opened 2021-03-08
feat: add tags to slates

akuokojnr:
This PR is a fix for issue #586. This is the first iteration for tagging in Slate. This PR creates and adds a tag input component to slate sidebar settings, slate object carousel and slate data object carousel. I have attached a video to show it works.

https://user-images.githubusercontent.com/31008944/111635456-65d01000-87ef-11eb-92ad-8246cd4fd58c.mp4


components/core/CarouselSidebarData.js
```diff
@@ -458,6 +466,23 @@ class CarouselSidebarData extends React.Component {
...
+    if (response.success) {
+      this.setState({ suggestions: response.tags });
+    }
+
+    if (Events.hasError(response)) {
+      return;
+    }
+  };
+
+  fetchSuggestions = async () => {
```
martinalong, line 481: One thing I might suggest for a future iteration of this is: instead of fetching these every time a user opens carousel sidebar data, we can instead fetch them all at once at the time of login (see viewer.js in getById). That way you can access it like this.props.viewer.tags

Or we could fetch it like how you do now (when the first open the editing panel) and then save it to viewer. Then after that if you see that viewer.tags already exists, you don't need to fetch again. That also has a similar effect (let me know if you want more details on how to do that)

martinalong: actually yeah do you think you can make one of those changes before I merge this PR? I noticed that the loading of tags can be a bit slow to do each time 
image

pages/api/tags/get-all-tags.js
```diff
@@ -0,0 +1,46 @@
...
+  const response = await Data.getSlatesByUserId({ userId: id });
+  if (!response) {
+    return res.status(404).send({ decorator: "SERVER_GET_SLATES_NOT_FOUND", error: true });
+  }
+
+  if (response.error) {
+    return res.status(500).send({ decorator: "SERVER_GET_SLATES_NOT_FOUND", error: true });
+  }
+
+  const filteredTags = response
```
martinalong, line 37: Also, post-database migration I have an optimization that will make this a lot faster, but for now we can leave it like this. I also notice that this doesn't get file level tags. Is that intentional? (getting file level tags will also be easier post migration)

pages/api/tags/get-all-tags.js
```diff
@@ -0,0 +1,46 @@
+import * as Utilities from "~/node_common/utilities";
```
martinalong, line 1: mind renaming this file to get-user-tags? I think in the future it's possible we'll have a separate get all tags (like get all tags across all of slate) and this might be confused for that

pages/api/tags/delete-tag.js
```diff
@@ -0,0 +1,82 @@
+import * as Utilities from "~/node_common/utilities";
+import * as Data from "~/node_common/data";
+import * as ViewerManager from "~/node_common/managers/viewer";
+import * as SearchManager from "~/node_common/managers/search";
+
+export default async (req, res) => {
+  const id = Utilities.getIdFromCookie(req);
+  if (!id) {
+    return res.status(500).send({ decorator: "SERVER_GET_SLATES_TAGS", error: true });
```
martinalong, line 9: Just a heads up: for decorator naming conventions, we're moving to a system where if they're not authenticated, it's called "SERVER_NOT_AUTHENTICATED" (so that we can have a single error message for all of this same error across different endpoints)

pages/api/tags/delete-tag.js
```diff
@@ -0,0 +1,82 @@
...
+  }
+
+  const user = await Data.getUserById({
+    id,
+  });
+
+  if (!user) {
+    return res.status(404).send({
+      decorator: "SERVER_GET_SLATE_USER_NOT_FOUND",
```
martinalong, line 18: And all other decorators will be "SERVER" + function name + error name 
Here it would be something like "SERVER_GET_USER_TAGS_USER_NOT_FOUND"
And "SERVER_GET_USER_TAGS_SLATES_NOT_FOUND" for when you are trying to getSlatesByUserId below

For every new error decorator, we then go into messages.js and add a coresponding error message to show the user (this is the message that appears in the red alert on the page when something goes wrong)

martinalong: Totally don't expect you to know this already since it isn't in the codebase yet (it's something I'm standardizing during the migration), but just wanted to give you a heads up about the upcoming change

components/core/CarouselSidebarData.js
```diff
@@ -458,6 +466,23 @@ class CarouselSidebarData extends React.Component {
...
+    const res = await Actions.getTagsByUserId();
```
martinalong, line 482: Actually, I think this is loading forever for me and never finishing. Not sure why. I just see the loader spinner. Also would be nice to have an option to add a tag by typing even before the list of tags loads

components/core/CarouselSidebarData.js
martinalong: Yeah I noticed that if you are going through the carousel, every time you go to the next item it calls this "get all tags" again, which can be pretty inefficient. I think it only makes sense to call it once you click into the tags input field, if you were to fetch upon first use of edit tags, rather than when the user first logs in

akuokojnr: Yeah Martina. You're very right. Thank you for pointing this out.

pages/api/tags/get-all-tags.js
akuokojnr: I decided to remove this endpoint altogether because it is not needed now. 

Also, I made a mistake by not getting file level tags. This has been fixed now.

pages/api/tags/delete-tag.js
akuokojnr: Got it!

pages/api/tags/get-all-tags.js
akuokojnr: This endpoint is not need again

pages/api/tags/delete-tag.js
akuokojnr: Thank you. I am going to add the corresponding messages

components/core/CarouselSidebarData.js
akuokojnr: Now, you can press "Enter" key to add a tag

akuokojnr:
@martinalong per my testing, all issues seem to be fixed. The only one outstanding is submit event handler in the SidebarEditTags (https://github.com/filecoin-project/slate/blob/%40akuokojnr/tags/components/sidebars/SidebarEditTags.js#L60). 

The issue I am encountering but I don't seem to find a fix when I submit bulk tags to be added to the data items, they are added to some of the items but not others. Same with removal. I can't seem to find what I am doing wrongly. I would love some insight on this. Thanks

martinalong:
Do you mind removing the delete tags functionality like we talked about in the meeting? I don't see any use cases where people will really need to use it, and it still has bugs in it. Like for example I deleted a tag (using the trash icon), and now I see the tag here

martinalong:
One of the issues I'm seeing is the same issue as before but in a different place, where if this.props.sidebarData.suggestions is null and you try to spread it, it'll error
image

I'd recommend going through your code and making sure that anywhere you use the spread operator, you either check before you spread the value, or you set a default value of [] for cases where it is null
```
if (this.props.sidebarData) {
    //code that spreads sidebarData
}
```

or

```
let sidebarData = this.props.sidebarData || []
//code that spreads sidebarData
```

components/sidebars/SidebarEditTags.js
```diff
@@ -0,0 +1,160 @@
...
+
+export default class SidebarEditTags extends React.Component {
+  state = {
+    tags:
+      !Array.isArray(this.props.sidebarData.commonTags) ||
+      this.props.sidebarData.commonTags?.length === 0
+        ? []
+        : this.props.sidebarData.commonTags,
+    suggestions:
+      !Array.isArray(this.props.sidebarData.suggestions) ||
```
martinalong, line 29: A cleaner way of doing this is:
```
state = {
    tags: this.props.sidebarData?.commonTags || []
}
```

The question mark after the sidebarData is just in case sidebar data is null

components/system/components/Tag.js
```diff
@@ -0,0 +1,581 @@
...
+  display: flex;
+  flex-wrap: wrap;
+`;
+
+const STYLES_INPUT_CONTAINER = css`
+  width: 100%;
+  position: relative;
+`;
+
+const STYLES_DROPDOWN = css`
```
martinalong, line 21: Mind rounding the corners on the dropdown to match the rest of the UI elements (using border radius)? It looks a little off when it's square corners when everything else is rounded.

components/system/components/Tag.js
```diff
@@ -0,0 +1,581 @@
+import * as React from "react";
```
martinalong, line 1: Also, there should be restrictions on what type of tags can be added. You should check/modify them both while typing and before adding them. Tags should be converted to lowercase, and only contain letters, numbers, or dashes. No spaces or unusual characters. 

If someone tries to type a non allowed character, they shouldn't be allowed to type it into the input (nothing will show up)
And if they type an uppercase character it'll be automatically converted to lowercase
And a space will automatically be converted to a dash

You can achieve this by controlling the input in the onChange function for the input

components/sidebars/SidebarEditTags.js
```diff
@@ -0,0 +1,160 @@
...
+  };
+
+  _handleSubmit = async () => {
+    this.props.onCancel();
+
+    let { checked, objects, commonTags } = this.props.sidebarData;
+    const checkedIndexes = Object.keys(checked);
+    let data;
+
+    await Promise.all(
```
martinalong, line 67: I suspect the reason some are updating and not others is becuase you are doing a promise.all (aka make all these calls to the server at once). Because browsers are single threaded (only do one thing at a time), trying to do this means only some go through and others don't.

What you should be doing (which is much more efficient) is making a new endpoint for the api which can handle updating multiple files at once and then just make a single call to that. You can make a new one and base it off the existing api/data/update.js

components/system/components/Tag.js
martinalong: Note: edited to change from hyphens to spaces. Spoke to haris and we decided that spaces look stylistically better than hyphens in tags

components/sidebars/SidebarEditTags.js
akuokojnr: Thank you so much for this!! I decided to modify `api/data/update.js` to handle an array of updates.

components/system/components/Tag.js
```diff
@@ -0,0 +1,542 @@
...
+      onChange({ target: { name: "tags", value: [...tags, value.trim()] } });
+
+      setValue("");
+    }
+  };
+
+  const _handleChange = (e) => setValue(e.target.value.toLowerCase());
+
+  const _handleKeyPress = (e) => {
+    let regex = /[a-z0-9\s-]/i;
```
martinalong: We decided to not have hyphens in tags, so mind changing this to `[a-z0-9\s]` instead of `[a-z0-9\s-]`?

components/system/components/Tag.js
```diff
@@ -0,0 +1,542 @@
...
+    }
+
+    if (onChange) {
+      onChange({ target: { name: "tags", value: [...tags, value.trim()] } });
+
+      setValue("");
+    }
+  };
+
+  const _handleChange = (e) => setValue(e.target.value.toLowerCase());
```
martinalong, line 475: Instead of using handle key press, I'd recommend handling stripping out blocked characters in the on change method. This is because if you paste something into the input field, it can get around the handle key press check (try pasting something like *$@#&*$&*! and it'll let you)

akuokojnr: Done!

PR #627, opened 2021-03-22
refactored codeblock to use prism-react

toastts:
it's backwards compatible w/older references to CodeBlock and I made the theme more or less the same

codeBlock


toastts:
there's some small text offsetting issues, I'll fix it after I test it some more and work out the bugs

PR #630, opened 2021-03-26
increased textile storage cap for users (yay!)

toastts:
made it 30GB, left the 50GB on email verification message in the data meter assuming that's still true


jimmylee:
Glad you discovered how to do this :-)

PR #682, opened 2021-03-30
disallow action bar when viewing other's profile

tarafanlin:
**disable this**

Screen Shot 2021-03-30 at 1 51 04 PM

### DIFF
--- a/components/core/DataView.js
+++ b/components/core/DataView.js
@@ -418,12 +418,12 @@ export default class DataView extends React.Component {
     });
   };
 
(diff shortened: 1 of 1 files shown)


martinalong (commented):
Instead of passing in isOwner to the functions, you can just check this.props.isOwner directly from the functions

tarafanlin:
> Instead of passing in isOwner to the functions, you can just check this.props.isOwner directly from the functions

fixed ^ @martinalong

PR #692, opened 2021-04-02
fix: carousel sidebar cover image

akuokojnr:
When looking at a file which I do not own, I can upload a cover image. This PR fixes it. 

See a screenshot of the issue below:

sidebar


martinalong:
Thanks for catching that!

PR #704, opened 2021-04-08
improvement: font object preview

Aminejvm:
Screen Shot 2021-04-08 at 8 01 30 PM


components/core/FontFrame/Views/FontObjectPreview.js
```diff
@@ -0,0 +1,50 @@
...
+
+const FontObjectPreview = React.memo(
+  ({ url, cid, fallback }) => {
+    const { isFontLoading, error, fontName } = useFont({ url, name: cid }, [cid]);
+    if (error) {
+      return fallback;
+    }
+    return (
+      <div style={{ fontFamily: fontName }}>
+        {isFontLoading ? <div>loading</div> : <div css={STYLES_LETTER}>Aa</div>}
```
martinalong, line 44: Mind using a different style for the loading state? Just "loading" looks a bit unfinished. Maybe we could use the same icon that fonts that don't load properly use:
image

martinalong:
Other than that one small nit, looks good!

PR #732, opened 2021-04-24
@tarafanlin/landingslate

tarafanlin:
temporary fix for landing page broken slate link, it's almost all managed account now so it should be fine for a while.
Screen Shot 2021-04-23 at 8 00 07 PM


jimmylee (approved):
LGTM

PR #742, opened 2021-05-05
Confirmation popup components

jasonleyser:
Created two new button styles:
- delete
- cancel

Created 3 popup modal components:
- single file, multi file and collection delete
- delete with input
- confirmation

Added a new validation rule for the input value to equal a string. Used to disabled and enable the delete button.

Screen Shot 2021-05-05 at 10 42 18 AM
Screen Shot 2021-05-05 at 10 52 39 AM

### DIFF
--- a/common/validations.js
+++ b/common/validations.js
@@ -193,3 +193,9 @@ export const isUnityFile = async (file) => {
     return false;
   }
 };
+
+export const isUsernameMatch = (data) => {
+  if(data.input == data.username) {
+    return true;
+  }
+}
--- a/common/validations.js
+++ b/common/validations.js
@@ -193,9 +193,3 @@ export const isUnityFile = async (file) => {
     return false;
   }
 };
-
-export const isUsernameMatch = (source, candidate) => {
-  if(source == candidate) {
-    return true;
-  }
-}
--- a/common/validations.js
+++ b/common/validations.js
@@ -194,8 +194,8 @@ export const isUnityFile = async (file) => {
   }
 };
 
-export const isUsernameMatch = (data) => {
-  if(data.input == data.username) {
+export const isUsernameMatch = (source, candidate) => {
+  if(source == candidate) {
     return true;
   }
 }
--- a/components/core/ConfirmationModal.js
+++ b/components/core/ConfirmationModal.js
@@ -9,7 +9,7 @@ import { ButtonPrimaryFull, ButtonDeleteFull, ButtonDeleteDisabledFull, ButtonCa
 import { Input } from "~/components/system/components/Input.js";
 
 const STYLES_TRANSPARENT_BG = css `
-  background-color: rgba(0,0,0,0.3);
+  background-color: ${Constants.system.bgBlurGrayBlack};
   width: 100%;
   height: 100vh;
   position: fixed;
@@ -21,7 +21,7 @@ const STYLES_TRANSPARENT_BG = css `
 const STYLES_MAIN_MODAL = css `
   width: 380px;
   height: auto;
-  background-color: #fff;
+  background-color: ${Constants.system.white};
   position: fixed;
   left: 50%;
   top: 50%;
@@ -32,22 +32,21 @@ const STYLES_MAIN_MODAL = css `
 `;
 
 const STYLES_HEADER = css `
-  color: #000;
-  font-size: 16px;
-  font-weight: bold;
-  font-family: 'inter-medium', -apple-system, BlinkMacSystemFont, arial, sans-serif;
+  color: ${Constants.system.black};
+  font-size: ${Constants.typescale.lvl1};
+  font-family: ${Constants.font.semiBold};
 `;
 
 const STYLES_SUB_HEADER = css `
-  color: #8E8E93;
-  font-size: 14px;
+  color: ${Constants.system.textGray};
+  font-size: ${Constants.typescale.lvl0};
   margin-top: 16px;
-  font-family: 'inter-regular', -apple-system, BlinkMacSystemFont, arial, sans-serif;
+  font-family: ${Constants.font.text};
 `;
 
 const STYLES_INPUT_HEADER = css `
-  color: #000;
-  font-size: 12px;
+  color: ${Constants.system.black};
+  font-size: ${Constants.typescale.lvlN1};
   font-weight: bold;
   margin-top: 24px;
   margin-bottom: 8px;
@@ -57,27 +56,19 @@ export const DeleteModal = (props) => {
     let header = '';
     let subHeader = '';
 
-    if(props.data.type == 'multi') {
-      header = 'Are you sure you want to delete the selected ' + props.data.selected + ' files?';
+    if (props.data.type === 'multi') {
+      header = `Are you sure you want to delete the selected ${props.data.selected} files?`;
       subHeader = 'These files will be deleted from all connected collections and your file library. You can’t undo this action.';
     }
 
-    if(props.data.type == 'single') {
-      header = 'Are you sure you want to delete the file "' + props.data.filename + '"?';
+    if (props.data.type === 'single') {
+      header = `Are you sure you want to delete the file "${props.data.filename}"?`;
       subHeader = 'This file will be deleted from all connected collections and your file library. You can’t undo this action.';
     }
 
-    if(props.data.type == 'collection') {
-      header = 'Are you sure you want to delete the collection “' + props.data.collection + '”?';
+    if (props.data.type === 'collection') {
+      header = `Are you sure you want to delete the collection "${props.data.collection}"?`;
       subHeader = 'This collection will be deleted but all your files will remain in your file library. You can’t undo this action.';
-    }
-
-    const _handleDelete = () => {
-      props.response(true);
-    }
-
-    const _handleCancel = () => {
-      props.response(false);
     }
 
     return (
@@ -86,8 +77,8 @@ export const DeleteModal = (props) => {
           <div>
             <div css={STYLES_HEADER}>{header}</div>
             <div css={STYLES_SUB_HEADER}>{subHeader}</div>
-            <ButtonCancelFull onClick={_handleCancel} style={{ margin: '24px 0px 8px' }}>Cancel</ButtonCancelFull>
-            <ButtonDeleteFull onClick={_handleDelete}>Delete</ButtonDeleteFull>
+            <ButtonCancelFull onClick={() => props.response(false)} style={{ margin: '24px 0px 8px' }}>Cancel</ButtonCancelFull>
+            <ButtonDeleteFull onClick={() => props.response(true)}>Delete</ButtonDeleteFull>
           </div>
         </div>
       </div>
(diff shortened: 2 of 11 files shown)


common/validations.js
```diff
@@ -193,3 +193,9 @@ export const isUnityFile = async (file) => {
     return false;
   }
 };
+
+export const isUsernameMatch = (data) => {
+  if(data.input == data.username) {
```
jimmylee, line 198: Use `===`

Also this function would be unusable without passing in an object, why not just make it a proper string compare function and do `export const isUsernameMatch = (source, candidate)`

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+  position: fixed;
+  left: 0;
+  top: 0;
+  z-index: 999999;
+`;
+
+const STYLES_MAIN_MODAL = css `
+  width: 380px;
+  height: auto;
+  background-color: #fff;
```
jimmylee, line 24: Constants seem appropriate here? cc: @martinalong

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+  font-weight: bold;
+  margin-top: 24px;
+  margin-bottom: 8px;
+`;
+
+export const DeleteModal = (props) => {
+    let header = '';
+    let subHeader = '';
+
+    if(props.data.type == 'multi') {
```
jimmylee, line 60: Whitespace missing `if(`

Is prettier enabled?

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+      header = 'Are you sure you want to delete the selected ' + props.data.selected + ' files?';
```
jimmylee, line 61: Use string template `Some words ${yourVariable}`

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+      subHeader = 'These files will be deleted from all connected collections and your file library. You can’t undo this action.';
+    }
+
+    if(props.data.type == 'single') {
```
jimmylee, line 65: Use `===` comparison

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+      header = 'Are you sure you want to delete the file "' + props.data.filename + '"?';
+      subHeader = 'This file will be deleted from all connected collections and your file library. You can’t undo this action.';
+    }
+
+    if(props.data.type == 'collection') {
+      header = 'Are you sure you want to delete the collection “' + props.data.collection + '”?';
+      subHeader = 'This collection will be deleted but all your files will remain in your file library. You can’t undo this action.';
+    }
+
+    const _handleDelete = () => {
```
jimmylee, line 75: You should just delete this and do `onClick={() => props.response()}`

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+      props.response(true);
+    }
+
+    const _handleCancel = () => {
+      props.response(false);
```
jimmylee, line 80: `onClick={() => props.response(false)}`

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+            <ButtonDeleteFull onClick={_handleDelete}>Delete</ButtonDeleteFull>
+          </div>
+        </div>
+      </div>
+    );
+};
+
+export const DeleteWithInputModal = (props) => {
+    const [isUsernameMatch, setIsUsernameMatch] = useState(false);
+    const header = 'Are you sure you want to delete your account @' + props.data.username + '?';
```
jimmylee, line 99: Use string template Some words ${yourVariable}

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+    const subHeader = 'You will lose all your files and collections. You can’t undo this action.';
+    const inputHeader = 'Please type your username to confirm';
+    const inputPlaceholder = 'username';
+
+    const _handleDelete = () => {
+      props.response(true);
+    }
+
+    const _handleCancel = () => {
+      props.response(false);
```
jimmylee, line 109: onClick={() => props.response(false)} on the element

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+
+export const DeleteWithInputModal = (props) => {
+    const [isUsernameMatch, setIsUsernameMatch] = useState(false);
+    const header = 'Are you sure you want to delete your account @' + props.data.username + '?';
+    const subHeader = 'You will lose all your files and collections. You can’t undo this action.';
+    const inputHeader = 'Please type your username to confirm';
+    const inputPlaceholder = 'username';
+
+    const _handleDelete = () => {
+      props.response(true);
```
jimmylee, line 105: See above

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+
+    const _handleOnChange = (e) => {
+     if(Validations.isUsernameMatch({ input: e.target.value, username: props.data.username })) {
+      setIsUsernameMatch(true)
+     }else{
+      setIsUsernameMatch(false)
+     }
+    }
+
+    let deleteButton;
```
jimmylee, line 120: let deleteButton = <ButtonDeleteDisabledFull>Delete</ButtonDeleteDisabledFull>;
    if (isUsernameMatch) {
      deleteButton = <ButtonDeleteFull onClick={_handleDelete}>Delete</ButtonDeleteFull>;
    }

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+
+    const _handleDelete = () => {
+      props.response(true);
+    }
+
+    const _handleCancel = () => {
+      props.response(false);
+    }
+
+    const _handleOnChange = (e) => {
```
jimmylee, line 112:
```
const match = Validations.isUsernameMatch({ input: e.target.value, username: props.data.username });
setIsUsernameMatch(match);
```

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+            <ButtonCancelFull onClick={_handleCancel} style={{ margin: '16px 0px 8px' }}>Cancel</ButtonCancelFull>
+            {deleteButton}
+          </div>
+        </div>
+      </div>
+    );
+};
+
+export const ConfirmModal = (props) => {
+    const header = 'Making this file private will remove it from the following public slates: ' + props.data.username + '. Do you wish to continue?';
```
jimmylee, line 144: Use string template Some words ${yourVariable}

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,167 @@
...
+    const subHeader = 'Making the file private means it’s not visible to others unless they have the link.';
+
+    const _handleConfirm = () => {
+      props.response(true);
+    }
+
+    const _handleCancel = () => {
+      props.response(false);
```
jimmylee, line 152: onClick={() => props.response(false)} on the element

components/system/components/Buttons.js
```diff
@@ -306,3 +306,143 @@ export const ButtonWarning = (props) => {
     />
   );
 };
+
+
+const STYLES_BUTTON_DELETE = css`
+  ${STYLES_BUTTON}
+  cursor: pointer;
+  color: ${Constants.system.white};
```
jimmylee, line 314: Constants are used here so thats nice, so using constants above will be useful.

jimmylee (changes requested):
Did a review to help out (cc: @martinalong)

common/validations.js
```diff
@@ -193,3 +193,9 @@ export const isUnityFile = async (file) => {
     return false;
   }
 };
+
+export const isUsernameMatch = (source, candidate) => {
```
martinalong, line 197: I'm usually all for breaking down things into more manageable functions, but I think this is one case where it's unnecessary and just adds more code and confusion. This method is essentially just `a === b` and should be left as that in the code rather than making this extra function which people would have to track down to see what it is

jasonleyser: okay sounds good, ill just move it to the component

components/system/components/Buttons.js
```diff
@@ -306,3 +306,143 @@ export const ButtonWarning = (props) => {
     />
   );
 };
+
+
+const STYLES_BUTTON_DELETE = css`
```
martinalong, line 311: I think we already have a similar button called ButtonWarning. See if you can use that instead of creating a one-off button. Also so that we can keep styles consistent across the UI

components/system/components/Buttons.js
```diff
@@ -306,3 +306,143 @@ export const ButtonWarning = (props) => {
...
+`;
+
+const STYLES_BUTTON_CANCEL_TRANSPARENT = css`
+  ${STYLES_BUTTON}
+  cursor: pointer;
+  background-color: transparent;
+  color: ${Constants.system.darkGray};
+`;
+
+export const ButtonCancel = (props) => {
```
martinalong, line 435: Same goes for here. Let's use ButtonSecondary instead of this for now but leave the code here for future use if we convert over

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,137 @@
...
+
+const STYLES_INPUT_HEADER = css `
+  color: ${Constants.system.black};
+  font-size: ${Constants.typescale.lvlN1};
+  font-weight: bold;
+  margin-top: 24px;
+  margin-bottom: 8px;
+`;
+
+export const DeleteModal = (props) => {
```
martinalong, line 55: To make this function more flexible, I think a better paradigm is to think of it as something where you pass in the header and subheader, and it renders those. Rather than storing all the headers and subheaders inside the component. Because imagine if we start using this a lot, we'd have to keep adding more and more to it for the various options.

Using this component should look something like: 
``` 
<DeleteModal 
    header={`Are you sure you want to delete the selected ${props.data.selected} files?`}
    subHeader={'These files will be deleted from all connected collections and your file library. You can’t undo this action.'}
/>
```

And for the specific case where you need to validate your username, you can just add a flag for that:
``` 
<DeleteModal 
    header={`Are you sure you want to delete your account?`}
    subHeader={'Type your username to continue'}
    inputValidation=true
    matchValue={this.props.viewer.username}
/>
```

martinalong: And on top of that, I don't think you need 3 separate modals for Confirm,Delete, and DeletewithInput. These can all be the same component, but with a flag like "type" that determines which style it is.

```
<ConfirmationModal
    type={"DELETE"} // I would recommend all caps for types. makes it so people don't have to think about capitalization
/>
```

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,137 @@
...
+            <div css={STYLES_SUB_HEADER}>{subHeader}</div>
+            <ButtonCancelFull onClick={() => props.response(false)} style={{ margin: '24px 0px 8px' }}>Cancel</ButtonCancelFull>
+            <ButtonDeleteFull onClick={() => props.response(true)}>Delete</ButtonDeleteFull>
+          </div>
+        </div>
+      </div>
+    );
+};
+
+export const DeleteWithInputModal = (props) => {
```
martinalong, line 88: Also, one thing I would add is using the Boundary component. This makes it so if someone clicks outside of the modal, you can automatically close the modal. See GlobalModal for an example of how it's used (and be sure to set enabled={true} on Boundary or it won't work)

components/core/ConfirmationModal.js
```diff
@@ -0,0 +1,137 @@
...
+
+    return (
+      <div css={STYLES_TRANSPARENT_BG}>
+        <div css={STYLES_MAIN_MODAL}>
+          <div>
+            <div css={STYLES_HEADER}>{header}</div>
+            <div css={STYLES_SUB_HEADER}>{subHeader}</div>
+            <div css={STYLES_INPUT_HEADER}>{inputHeader}</div>
+            <Input placeholder={inputPlaceholder} onChange={_handleOnChange} />
+            <ButtonCancelFull onClick={() => props.response(false)} style={{ margin: '16px 0px 8px' }}>Cancel</ButtonCancelFull>
```
martinalong, line 113: The standard naming structure for actions is: "on[Action]". so like onChange, onConfirm, onSelect, etc which correspond to a _handleAction, _handleChange, _handleConfirm... in the parent. Staying consistent with this just makes it easier to know which props are functions and which ones are values

martinalong:
Yeah I think after those changes we should be good! Could you go through the codebase and swap window.confirm, window.prompt, etc out for this new component?

jasonleyser:
@martinalong should be good now

Screen Shot 2021-05-11 at 9 20 35 AM

Screen Shot 2021-05-11 at 7 33 08 PM

Screen Shot 2021-05-12 at 8 24 02 AM

PR #763, opened 2021-05-25
authentication with emails, twitter, and a new password strat

toastts:
check out https://www.notion.so/Slate-Auth-ea0982503ccf4cfcb804d397fc188062, it's the doc explaining most of the stuff in this PR

looks and feels pretty rad to use


martinalong:
image
Minor warning I get in the terminal^

martinalong:
image
when i click the proceed arrow on this, it shouldn't let me continue to the next page. it should instead error and say something like "please enter a valid email or username" (either inside the login panel, or using the alert at the top)

martinalong:
image
Is it possible to make it so you don't have to authorize slate to access your info every time you sign in with twitter? I think other sites manage to do it only once during sign-up

And also can we change the name slate-test to Slate and add the logo as the photo?

martinalong:
Login situations where it doesn't behave as expected:
- Case: I signed up with twitter, then try to sign up using the email associated with that twitter account. It detects that that email exists, sends me to the sign-in page to enter my password instead (which won't work b/c i don't have a password). 
    - Instead, we should detect that they have a twitter account, and either error, or send them directly to sign-in with twitter
- case: I signed up wtih twitter, then try and sign-in with username or email. It proceeds to let me try and enter a password, then errors (because I have none)
    - Instead (if possible) can we either error and say "you previously signed in with twitter" or send me directly to sign in with twitter?
- Case: i signed upw ith email, then try to sign up again using the same email. It currently allows me to proceed to enter a password
    - It should error when i hit "send verification link" and tell me that an account with that email already exists, and not let me proceed
    
I'll continue adding to this list tomorrow

martinalong:
Continuing the previous list:
- If someone has a legacy account (no email) and tries to log in, it should block them from using slate until they enter an email on the next page and verify it. This is so we can convert all the old users to the new format. Right now it just lets me log in as usual
-When someone tries to sign up iwth an email/username and there is already an account associated with it, the sign-in page they then get sent to should have a message saying "There is already an account associated with your email" (just so that it's clear they are being sent to sign in instead of sign up)

martinalong:
image
And also if a code isn't 6 characters and you try to hit forward, nothing happens. There should be an error of some kind

martinalong:
So I've already caught a lot of the edge cases by just going through the sign in flow in unusual ways. @Aminejvm @toastts if you guys could go through the flow and test for edge cases yourselves, that would be great! Would probably be faster than me testing them then writing them out here

Look for places where there should be an error/message but isn't one, where there should be some sort of validation or confirmation before progressing, where you are allowed to progress when you shouldn't be, and where things may be confusing and may need clarification (like stating the password requirements)

martinalong:
Let's use a custom design or the existing alert system for this warning instead of the built in system which doesn't match our design style:

image

PR #785, opened 2021-06-11
Fix: Update password-reset template

Aminejvm:


components/core/Field.js
```diff
@@ -105,22 +97,20 @@ export default function Field({
...
   return (
     <div>
       <ContainerComponent>
-        <Input inputCss={STYLES} {...props} />
+        <Input inputCss={[STYLES_INPUT, STYLES_STATUS]} {...props} />
       </ContainerComponent>
-      {props.name === "password" && validations ? (
-        <PasswordValidations validations={validations} />
-      ) : (
+      {props.name === "password" && validations(<PasswordValidations validations={validations} />)}
```
martinalong, line 112: Getting this error with this line when I try to sign in 
image

Aminejvm: Thank you @martinalong it should be fixed now

PR #802, opened 2021-06-24
Intercom user data

jasonleyser:
Fixed bug where the user data wouldn't pass to the Intercom popup.

Now the intercom app is updated first with the user data, and then opened.

### DIFF
--- a/components/core/ApplicationUserControls.js
+++ b/components/core/ApplicationUserControls.js
@@ -103,15 +103,20 @@ const STYLES_ITEM_BOX = css`
 `;
 
 const OpenIntercom = ({ user }) => {
-  const { show } = useIntercom();
-
-  const showWithProps = () =>
-    show({
-      name: user.data.name || user.username,
-    });
+  const { show, update } = useIntercom();
 
   return (
-    <span style={{ cursor: "pointer", display: "block" }} onClick={() => showWithProps()}>
+    <span
+      style={{ cursor: "pointer" }}
+      onClick={() => {
+        update({
+          name: user.data.name,
+          username: user.username,
+          email: user.email,
+        });
+        show();
+      }}
+    >
       Help
     </span>
   );
--- a/components/core/ApplicationUserControls.js
+++ b/components/core/ApplicationUserControls.js
@@ -102,7 +102,7 @@ const STYLES_ITEM_BOX = css`
   }
 `;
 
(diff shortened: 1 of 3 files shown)


components/core/ApplicationUserControls.js
```diff
@@ -103,14 +103,21 @@ const STYLES_ITEM_BOX = css`
...
-  });
-
   return(
-    <span style={{ cursor: "pointer" }} onClick={() => showWithProps()}>
+    <span 
+      style={{ cursor: "pointer" }} 
+      onClick={() => {
+          update({
+            name: user.data.name,
+            username: user.username,
```
martinalong, line 114: Also, it looks like username is not being passsed through properly.  Can you make sure username and ideally the user's id is passed through? (not to be mistakened for the id that intercom generates for the user) 

I already added it on the intercom side as a custom data attribute like this, you just need to figure out how to pass them in: https://www.intercom.com/help/en/articles/179-send-custom-user-attributes-to-intercom
image
image

components/core/ApplicationUserControls.js
```diff
@@ -103,14 +103,21 @@ const STYLES_ITEM_BOX = css`
 `;
 
 const OpenIntercom = ({ user }) => {
```
martinalong, line 105: You should pass in onTogglePopup from the parent component and call it upon click to close the dropdown menu upon selecting "help"

```
<OpenIntercom
  user={this.props.viewer}
  onTogglePopup={this.props.onTogglePopup}
/>

```

```
onClick={() => {
  onTogglePopup();
  show(...);
}}

```

martinalong, line 105: I made this fix myself b/c it was small, but just a heads up that you needed to destructure it here, otherwise it won't exist:

`const OpenIntercom = ({ user, onTogglePopup }) => {`

martinalong:
It looks like the way that intercom is persisting your "chat history" is through an intercom session key, which isn't available when you log in via incognito mode, so if I log onto slate in incognito mode (or through a different browser or through mobile), my intercom chat history won't be there. Also if you log into two different accounts, you'll have the same intercom profile for both. 

Do you mind checking to see if there's other ways to persist this key that intercom offers? Like saving it to our database and then pasing it in? If they don't offer that option, it's fine.

The intercom session key:
image

Normally, message history looks like:
image

In incognito mode, message history is missing:
image

PR #835, opened 2021-07-13
Feat: Activity Page

Aminejvm:
This PR adds the following features

* New Object previews
* New Collection previews
* New Profile preview
* Updated Activity scene


martinalong:
image
image


For the markdown files, it looks like the preview is showing the code for it (?) and you can see that the actual interpreted text is different in the full size view. Is it possible to display the actual interpreted text in the preview? We may have to use something similar to the markdown renderer component to do so

components/core/ObjectPreview/3dObjectPreview.js
```diff
@@ -0,0 +1,38 @@
+import "isomorphic-fetch";
+
+import * as React from "react";
+import * as Styles from "~/common/styles";
+
+import { P } from "~/components/system";
+import { css } from "@emotion/react";
+
+import ObjectPreviewPremitive from "./ObjectPreviewPremitive";
```
martinalong, line 9: Misspelling: Premitive -> Primitive

martinalong:
Also can you make sure that the style for empty collections is the same as the style for nonempty ones?
image

common/activity-utilities.js
```diff
@@ -0,0 +1,175 @@
+/* eslint-disable no-prototype-builtins */
+import shuffle from "lodash/shuffle";
+import * as Actions from "~/common/actions";
+
+// NOTE(amine): fetch explore objects
+export const fetchExploreItems = async ({ currentItems, update }) => {
+  const requestObject = {};
+  if (currentItems.length) {
+    if (update) {
+      requestObject.latestTimestamp = currentItems[0].createdAt;
```
martinalong, line 10: Things aren't necessarily ordered by date anymore, so this method won't work. In order to get the latest/earliest timestamp you'll need to search through all the activity items

components/core/ActivityGroup/components/ProfileInfo.js
```diff
@@ -0,0 +1,101 @@
...
+export default function ProfileInfo({ owner, viewer, time, action, onAction }) {
+  const { isFollowing, handleFollow } = useFollowProfileHandler({ viewer, user: owner, onAction });
+  const { username, data = {} } = owner;
+  const { photo } = data;
+
+  const isOwner = viewer?.id === owner.id;
+  return (
+    <Link href={`/$/user/${owner.id}`} onAction={onAction}>
+      <div css={STYLES_PROFILE_CONTAINER}>
+        <img src={photo} alt={`${username} profile`} css={STYLES_PROFILE} />
```
martinalong, line 51: Instead of this text as alt, can we use some fallback image or just a grey background as a fallback image? I think the text looks really awkward

Could do something like this:
```
<img src="original-image.jpg"
   onerror="this.onerror=null;this.src='default-image.jpg';"
>
```

image

common/utilities.js
```diff
@@ -66,3 +66,42 @@ export const coerceToArray = (input) => {
...
+
+  const differenceInDays = Math.floor(differenceInHours / 24);
+  if (differenceInDays < 24) {
+    return differenceInDays + "d";
+  }
+
+  const currentYear = now.getFullYear();
+
+  const day = pastDate.getDay();
+  const month = pastDate.toLocaleString("default", { month: "long" });
```
martinalong, line 99: Can we display the date in the style that's in the designs? With a shortened month and in the "Month Day, Year" format like Aug 5, 2020

components/system/components/Typography.js
```diff
@@ -18,6 +18,31 @@ const LINK_STYLES = `
...
+    },
+    [color]
+  );
+
+const truncateElements = (nbrOfLines) =>
+  nbrOfLines &&
+  css`
+    overflow: hidden;
+    line-height: 1.5;
+    word-break: break-all;
```
martinalong, line 39: I'd recommend using word-break: break-all and overflow-wrap: break-word instead. break-all breaks things off in the middle of a word which looks strange

common/activity-utilities.js
Aminejvm: Here I'm not using the groups, currentItems are the items from getActivity/getExplore response before the grouping happens. I assume those are ordered by date

martinalong: Ahh ok yep that's fine then

martinalong:
There needs to be more spacing above the description for link objects (it's a bit squished compared to the file objects)
image

martinalong:
I think the bottom section should always have a top border, whether expanded or not, like in the second image. B/c oftentimes the colors are too similar to the file preview

image

image

martinalong:
I'm going to merge it for now b/c everyone really wants to try it out on dev. I'll keep adding notes here as I review though

PR #836, opened 2021-07-13
added new profile avatars

jasonleyser:
Replaced the old avatar components with a new ProfilePhoto component. The component has logic to use a BoringAvatar if the user hasn't uploaded a photo yet, using the user id to generate an design.

Also updated the user creation scripts to now leave the profile photo blank at the point of creation.

### DIFF
--- a/components/core/ApplicationUserControls.js
+++ b/components/core/ApplicationUserControls.js
@@ -6,6 +6,7 @@ import * as UserBehaviors from "~/common/user-behaviors";
 import { PopoverNavigation } from "~/components/system";
 import { css } from "@emotion/react";
 import { Link } from "~/components/core/Link";
+import ProfilePhoto from "~/components/core/ProfilePhoto"; 
 
 import { Boundary } from "~/components/system/components/fragments/Boundary";
 
@@ -118,16 +119,14 @@ export class ApplicationUserControlsPopup extends React.Component {
     if (this.props.popup === "profile") {
       const topSection = (
         <div css={Styles.HORIZONTAL_CONTAINER} style={{ marginBottom: 14 }}>
-          <span
-            css={STYLES_PROFILE_IMAGE}
-            style={{
-              cursor: "default",
-              width: 46,
-              height: 46,
-              marginRight: 16,
-              backgroundImage: `url('${this.props.viewer.data.photo}')`,
-            }}
-          />
+          <div style={{ marginRight: '16px', cursor: 'default' }} >
+            <ProfilePhoto 
+              size={46}
+              userId={this.props.viewer.id}
+              photo={this.props.viewer.data.photo} 
+            />
+          </div>
+
           <div
             css={Styles.VERTICAL_CONTAINER}
             style={{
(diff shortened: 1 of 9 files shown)


martinalong:
image
It looks like the profile page image component isn't using the new profile image component yet

martinalong:
image
Same thing with the directory page profile images

pages/api/twitter/signup-with-verification.js
```diff
@@ -103,18 +103,12 @@ export default async (req, res) => {
...
-      "https://slate.textile.io/ipfs/bafkreick3nscgixwfpq736forz7kzxvvhuej6kszevpsgmcubyhsx2pf7i",
-  });
-
   const user = await Data.createUser({
     username: newUsername,
     email: newEmail,
     twitterId: twitterUser.id_str,
     data: {
-      photo,
+      photo: "",
```
martinalong, line 111: I think it's best if we leave it out, rather than setting photo to an empty string

pages/api/twitter/signup.js
```diff
@@ -85,18 +85,12 @@ export default async (req, res) => {
...
-      "https://slate.textile.io/ipfs/bafkreick3nscgixwfpq736forz7kzxvvhuej6kszevpsgmcubyhsx2pf7i",
-  });
-
   const user = await Data.createUser({
     username: newUsername,
     email: newEmail,
     twitterId: twitterUser.id_str,
     data: {
-      photo,
+      photo: "",
```
martinalong, line 93: I think it's best if we leave it out, rather than setting photo to an empty string

components/core/ApplicationUserControls.js
```diff
@@ -118,16 +119,14 @@ export class ApplicationUserControlsPopup extends React.Component {
...
-              width: 46,
-              height: 46,
-              marginRight: 16,
-              backgroundImage: `url('${this.props.viewer.data.photo}')`,
-            }}
-          />
+          <div style={{ marginRight: '16px', cursor: 'default' }} >
+            <ProfilePhoto 
+              size={46}
+              userId={this.props.viewer.id}
```
martinalong, line 125: I think it's simpler if we just pass in a prop `user` rather than having to pick out the user id and user photo. Since there's always a user object and it's always structured the same way. The ProfilePhoto component can pick out user.id and user.data.photo out of it.

This way there's less to write each time (and less room to mess up), and there's fewer things to change if we decide to change how user is structured in the future

components/core/SearchModal.js
```diff
@@ -83,7 +84,12 @@ const UserEntry = ({ user }) => {
   return (
     <div css={STYLES_ENTRY}>
       <div css={STYLES_ENTRY_CONTAINER}>
-        <div style={{ backgroundImage: `url(${user.data.photo})` }} css={STYLES_PROFILE_PREVIEW} />
+        <div css={STYLES_PROFILE_PREVIEW}>
+          <ProfilePhoto 
+            user={this.props.viewer}
```
martinalong, line 89: I think you mean to pass in the user here not this.props.viewer
image

components/core/SearchModal.js
```diff
@@ -113,7 +119,14 @@ const STYLES_PROFILE_IMAGE = css`
 const UserPreview = ({ user }) => {
   return (
     <div>
-      <div css={STYLES_PROFILE_IMAGE} style={{ backgroundImage: `url('${user.data.photo}')` }} />
+      <div css={STYLES_PROFILE_IMAGE}>
+        <ProfilePhoto 
+          photo={user.data.photo}
```
martinalong, line 124: And you forgot to update this to the new format where you pass in user instead of photo and id

PR #851, opened 2021-08-04
Tweaks: Block Previews

Aminejvm:
This Pr fixes the following bugs
* Revoke new changes to the ObjectPreview's checkbox
* Add debounce on MouseMove to The Object Preview
* Show description on hover in CollectionPreview
* Fix 404 when visiting collections from the `subscribed` tab
* Add title prop to Previews


martinalong:
Looks good! I did notice this one issue though (not just with this branch, but with the new objects style in general) where link objects don't have the blue border when selected

image
image
image

PR #852, opened 2021-08-04
Chore: Use NEXT_PUBLIC format for env variables

### LINKED ISSUE #850: [CLEANUP] Convert env variable passing to browser to the NEXT_PUBLIC_ format
Props to amine for bringing up this improvement!

The way we currently pass Env variables to the browser is in server.js in each endpoint through the "resources" object. The NEXT_PUBLIC_ format is much easier to deal with. Let's convert the things we typically pass through resources to the NEXT_PUBLIC_ format (and make the appropriate name substitutions throughout the codebase)

Aminejvm:
This PR resolves #850

### DIFF
--- a/common/environment.js
+++ b/common/environment.js
@@ -1,5 +1,7 @@
+export const NODE_ENV = process.env.NEXT_PUBLIC_NODE_ENV;
+
 //NOTE(amine): feature flags
-export const ACTIVITY_FEATURE_FLAG = !!process.env.NEXT_PUBLIC_ACTIVITY_FEATURE_FLAG;
+export const ACTIVITY_FEATURE_FLAG = NODE_ENV === "development" || NODE_ENV === "staging";
 
 // NOTE(jim): External servers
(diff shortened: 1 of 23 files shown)


martinalong:
Looks great! Mind updating shovel, lens, and fiji as needed as well? Not because they need these variables to be front end variables, but so that we can keep consistent naming across the services, and just use one concise env file for each of them

common/environment.js
```diff
@@ -1,2 +1,7 @@
 //NOTE(amine): feature flags
 export const ACTIVITY_FEATURE_FLAG = !!process.env.NEXT_PUBLIC_ACTIVITY_FEATURE_FLAG;
```
martinalong, line 2: Can you update this to use NEXT_PUBLIC_NODE_ENV instead? And update the code that checks for it to check if NEXT_PUBLIC_NODE_ENV === "staging". The possible values for it will be development, staging, and production. It's already been added to the staging and production servers

PR #864, opened 2021-08-05
Update: Object and Collection previews tweaks

Aminejvm:
* [x] update stroke and shadow for Object/Collection Preview
* [x] update ratio of Object/Collection Preview
* [x] fallback assets for LinkObject
* [x] visit link from LinkObject
* [x] update Hover animations to use transforms
* [x] add useIsomorphicLayoutEffect to remove NextJs warnings
* [x] show a one line description on static Object/Collection Preview

### DIFF
--- a/common/constants.js
+++ b/common/constants.js
@@ -209,8 +209,8 @@ export const grids = {
     mobile: { width: 166, rowGap: 12 },
   },
   collection: {
-    desktop: { width: 432, rowGap: 16 },
-    mobile: { width: 300, rowGap: 8 },
+    desktop: { width: 382, rowGap: 16 },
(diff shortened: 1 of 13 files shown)


martinalong:
The description of this one seems to overlap the line below when popped up

image


And after this one popped up then retracted again, the full text seemed to still be there and overlapping 

image

martinalong:
Did you increase the time of the delay before the popup? It feels a bit non reponsive and delayed now compared to before

martinalong:
Going to merge this so tara can play around with the hover interaction and see what needs to be tweaked

martinalong:
Also it should show collection.data.name instead of collection.slatename:
image

PR #865, opened 2021-08-06
Update: Object and Collection previews tweaks

Aminejvm:


martinalong:
I think you may be checking the existence of the wrong property to determine whether to display a substitute graphic for links.

Previously, links had a coverImage value (which represented the screenshot of the page), but I've removed this since it didn't look very good and we are now just using the object.data.link.image url as the image. Check for this property instead, and if it exists show it

Here, the instagram link has just a broken image symbol instead of the graphic like it's supposed to.
And the elastic search link which does have an image shows the graphic instead. And same for the twitter link

image
image

PR #875, opened 2021-08-13
Fix: a bug where objects reappearing after deletion

### LINKED ISSUE #834: [BUG] In slate layout, things reappear after deleting sometimes even though deleted
If you try deleting something from a slate, it'll disappear then reappear and stick around even after it's been deleted sometimes. So some intermediary, incorrect update is happening it seems

Aminejvm:
forked from https://github.com/filecoin-project/slate/pull/872
This PR resolves #834


martinalong:
Looks great! So much nicer now that this is fixed, thanks for patching it

PR #887, opened 2021-08-19
Improvement: update Viewer's getById method

Aminejvm:
small improvement to speed up the app launch

### DIFF
--- a/node_common/managers/viewer.js
+++ b/node_common/managers/viewer.js
@@ -148,15 +148,15 @@ export const getById = async ({ id }) => {
 
   // user.library = await Data.getFilesByUserId({ id, sanitize: true });
 
-  const slates = await Data.getSlatesByUserId({
-    ownerId: id,
-    sanitize: true,
-    includeFiles: true,
-  });
-  const keys = await Data.getAPIKeysByUserId({ userId: id });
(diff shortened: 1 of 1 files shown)


node_common/managers/viewer.js
```diff
@@ -148,15 +148,15 @@ export const getById = async ({ id }) => {
...
-  const slates = await Data.getSlatesByUserId({
-    ownerId: id,
-    sanitize: true,
-    includeFiles: true,
-  });
-  const keys = await Data.getAPIKeysByUserId({ userId: id });
-  const subscriptions = await Data.getSubscriptionsByUserId({ ownerId: id });
-  const following = await Data.getFollowingByUserId({ ownerId: id });
-  const followers = await Data.getFollowersByUserId({ userId: id });
+  const [slates, keys, subscriptions, following, followers, { bucketRoot }] = await Promise.all([
```
martinalong, line 151: Great thinking executing them all at once! One thing I would suggest is maybe making this a Promise.allSettled instead, so that even if one fails, they won't all fail
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled

Aminejvm: Thank you! I didn't know about this, `allSettled` will fit perfectly for this issue

PR #895, opened 2021-08-24
Update: new top navigation

### LINKED ISSUE #894: New top bar style

Aminejvm:
This PR resolves #894 
* added new utility components (Switch & Show). Hopefully, it'll make our jsx more readable and easier to understand
* moved DataMeter to navigation
* remove filtering and the data meter from my slate
* very small update to search
* update navigation to new design
* 
One thing left is the extension button, I need to do a spike on how to detect each browser to show its compatible link. Also, I need to sync to Jason so we don't show the button to the users that already downloaded the extension.


martinalong:
image
The lower two divider lines seem to be doubled up and thicker than the top line

martinalong:
image
The cursor here makes it look like this is clickable when it isn't

PR #944, opened 2021-09-15
Update: Saving flow

Aminejvm:
This new saving flow uses a Queue system to manage uploads. It'll solve the conflicts caused by how Textile works. The new system also made it simple to retry uploading after a failure.

https://user-images.githubusercontent.com/39884652/133501758-11531637-89ca-49af-bf76-08dc9fb6deae.mov

We can also detect duplicates now and the user will be able to edit them

https://user-images.githubusercontent.com/39884652/133502100-03655530-a525-4174-8ab7-4efe89281881.mov

We can cancel all or individual uploads

https://user-images.githubusercontent.com/39884652/133502432-bc0c1d5c-a8a3-4059-bd2c-a6633a08824b.mov

**Thoughts on code**
For this feature, I tried to keep related code as close together as possible. I moved all the upload code from different components to `components/core/Upload` folder which contains now all the app's upload components. This will make it easier to debug and maintain.  For the upload logic, it's currently located in `file-utilities` and `upload-utilities`.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Update-Saving-flow-61151a37a89f4eac9bf8826035b28b75) by [Unito](https://www.unito.io)


martinalong:
Really fantastic work with the upload Provider and upload queue! You did a great job organizing things so clearly! 

The only comments I have are a few UX improvements:

- Can we make the objects in the upload summary clickable? So that clicking them would take you to the file in your library? To do so I think all you would need is the cid, then you could make it a link using the custom component we have
`<Link onAction={onAction} href="/_/data?cid=FILE-CID-HERE">{button}</Link>`

- When you're on the upload summary page and you drag in a file, it takes a couple seconds before it registers in the uploading box at the top. Is it possible to make this any faster? Or is there some limitation on it?

- Let's remove the library items from the upload summary. We decided not to have the full upload history--just the items from the current upload session (both completed and ongoing saves). I think loading all the library files is what's causing the long delay when you click "upload summary"

- I'm curious why these duplicate files remain at the top even when I upload new things (the two files below them) that are more recent? Was this a conscious choice? (if so that's fine) If not, maybe it makes more sense to order things by recency, with most recent at the top
image

- The upload queue works to prevent double file uploads, but you're still able to double-upload a link in quick succession
image

- Do links show up in the upload summary when they are saving (not just after they are saved)? If not, is it possible to add them?

I have a few more UX changes for Tara to make (related to the upload modal automatically opening, and maybe combining the upload and upload summary modals), but I'll wait till this PR is submitted so she can try it out herself and then decide how to implement it

Aminejvm:
Thank you for the `Link` advice! using it not only cleaned the implementation but also improved the speed of the summary page. Now it’s reasonably fast when switching to the summary page and when adding files to upload

For sorting, yeah that was intentional, but I agree it’s not intuitive. One thing that’s wrong with the most recent to the top, is that files will be uploaded from the bottom to the top since we’re using a queue. To fix that I ordered by status and date (failed files first, then files being saved then duplicates, and finally saved ones) let me know how you think about that, if it’s still confusing I’ll be happy to revert it to sorting by recency

For links, I’ve added them to the queue system, it should work fine similar files

martinalong:
image

The total count of objects (here it says 4) seems off sometimes when I "double upload" in quick succession. To achieve this, I uploaded two objects, then while they were uploading, dragged those two objects in again, and then the count doesn't match up. I dragged another file in after that (hence why there's a third).

martinalong:
Great job speeding up the UI reaction! And the new ordering schema you implemented looks great!

martinalong:
image
Can we make "objects" and "sizes" singular? Aka "object" and "size"

martinalong:
image
Also, for this zora link, I have not uploaded it before, but it shows as "already saved" because I uploaded the link twice in quick succession. I think in the case of rapid double uploads (where it is in the upload queue twice, but not in the library yet), it should just skip the duplicate automatically but still show "saving" for the object instead of "already saved", otherwise people will think it isn't uploading at all

PR #954, opened 2021-09-24
Filtering feature

Aminejvm:
This is an initial version, includes filtering by file types and link types

**Components structure**
It follows a similar structure to the uploading feature, most of the filtering logic is tied together. The filter component is broken into multiple small components, all rendered in the `SceneFilesFolder` except for the `Filter.Navbar` component which is rendered in a portal. This way we can avoid making it a global state and only render the elements that use the filter. Here is a representation of how the code translates to the UI
`Filter.Navbar` and `Filter.NavbarPortal` are using `PortalsProvider` to render the child elements, I found this pattern in this article [Avoid global state](https://jjenzz.com/avoid-global-state-colocate)

**Filtering resprensentation**
For now, I modeled the filtering process as views, subviews, and filters. Each view will have filters and maybe subviews. e.g. Browser Filtering is a view with `all` as a filter and  `history`, `bookmarks`, `saved` as subviews.
image
Each filter type has a handler mapped to it, which will be used in the web worker. 
I'm thinking of changing this structure when adding other filters, to a tree data structure. I feel like the current structure isn't intuitive, and maybe hard to use when adding new filters

**Filtering via web workers**
The filter keeps track of which filter to use by knowing its parent view and its type (for some cases also the subview). Each time one of those parameters changes, `useFilterWorker` will submit them to the web worker which in return will give us the results. One benifit we get from this is we won't get into racing conditions, only one web worker will be running (useFilterWorker will terminate the old Worker before spawning a new one )

┆Issue is synchronized with this [Notion page](https://www.notion.so/Filtering-feature-5063a81730064bbdafc9ec666e739ebb) by [Unito](https://www.unito.io)


common/filter-utilities.js
```diff
@@ -0,0 +1,86 @@
+import {
+  isImageType,
+  isVideoType,
+  isAudioType,
+  isDocument,
+  isTwitterLink,
+  isYoutubeLink,
+  isTwitchLink,
+  isGithubLink,
+  isInstagramLink,
```
martinalong, line 10: Just a heads up: in the future instead of presets, what we will do is take a user's top saved link domains, and list them. For now we can leave it like this as a preview to see if we want to keep this.

martinalong:
I'm going to merge this so Haris and Tara can try it out and see how they like it. Then we'll make revisions to the UI and the organization of the filters

PR #959, opened 2021-09-28
Additional changes for save

Aminejvm:
This PR solves #953 
* fix wrong total numbers
* combine upload and summary on the same page
* add a popup screen when dragging files into Slate


martinalong:
image
Hmm for some reason no files are uploading for me

martinalong:
image
I don't think we need the seconds for the time btw

martinalong:
I asked tara for some design tweaks as well. This isn't quite what I had in mind for the combined pages design. She'll get back to you with the edited designs later

martinalong:
image
Can we change this text to "Drag and drop files to save them to slate"? And if someone is on the page for a collection aka tag, it'll replace "slate" with the name of the tag ("Drag and drop files to save them to #dog-photos")

martinalong:
image
The scroll bar is overlapping the checkmarks. Can we make it thin style or shift the ui a bit so it doesn't overlap? I think we should keep the scroll bar visible though. It's helpful to have here

martinalong:
Also can we add a background blur and tint when the modal is open? I think this should happen for all of our modals. It looks messy without it. Check with tara if you need specfics on what the blur and tint should look like, but I think it should also already be in the constants file

martinalong:
image
Somehow hyperfast parrot caused it to be 102% uploaded XD
Let's cap the percentage at 100% just in case of unusual cases like this

martinalong:
Maybe we should also close the upload modal automatically after a user adds a link or chooses a file? I think most people will only upload one link at a time this way, or one set of files all at once so that'll fit most use cases better. But I'm open to leaving it how it is if you think that makes more sense. Up to you!

PR #972, opened 2021-10-27
Feat: New global carousel design

### LINKED ISSUE #948: Carousel view
┆Issue is synchronized with this [Notion page](https://www.notion.so/Carousel-view-36b9ec4fd7014343929c5cd848fc956d) by [Unito](https://www.unito.io)

Aminejvm:
This PR resolves #948 by updating visuals for the carousel in both desktop and mobile views. In addition, we also now have new jumpers to interact with the files instead of the sidebar.

┆Issue is synchronized with this [Notion page](https://www.notion.so/Feat-New-global-carousel-design-555ae8e9247c4cdaa827c608563e425f) by [Unito](https://www.unito.io)


martinalong:
For adding to a tag: 

Rather than reordering all the tags after the backend sends the update, we should save the value of tags locally (aka slates / collections) from props just upon loading, and then after don't update it as props updates. Then we won't have to deal with reordering (which I think is confusing for users also)

I also think the select/unselect should visually happen immediately, by just changing the value locally within that component. That way it is done instantaneously rather than with a slight delay.

```
state = {
    collections: props.collections
}
//use this.state.collections rather than this.props.collections, and it won't change even as server sends updates

```

Same goes for the slate creation issue. I think you can just "artificially" (and instantaneously, rather than waiting for the server response) create a tag that's selected. I also don't think the jumper should close upon creating a new tag.

martinalong:
image
Let's put a dismiss x in the top right of each of these jumpers, and a blurred out background behind the jumper (maybe not as heavily blurred as the upload modal though), and make it so you can exit by clicking anywhere outside it as well. Same as what we did with the upload

martinalong:
When someone is typing in the tag creation/search field, spaces should automatically be swapped with hypens "-", and any other not allowed characters should just be ignored. You may have to make a modified regex function of the existing one to do this. B/c currently ' and / are swapped with - automatically, when they should be skipped

martinalong:
Generally transitions/reactions seem to be a bit laggy with the carousel (when you hover over tags, it takes a second for them to transition). I'm not sure if this is because i'm working with bad internet or something else. Mind looking into speeding those up if it's not just me?
https://www.loom.com/share/71376ff6fa6b46c1b14e7537fc136fca

Aminejvm:
> Generally transitions/reactions seem to be a bit laggy with the carousel (when you hover over tags, it takes a second for them to transition). I'm not sure if this is because i'm working with bad internet or something else. Mind looking into speeding those up if it's not just me? https://www.loom.com/share/71376ff6fa6b46c1b14e7537fc136fca

This performance looks really terrifying. I had a suspicion that it is maybe caused when viewing a Gif, but still worked fine on my end. I'll try to test it on other machines to see if I can replicate this performance

martinalong:
image
The icon and text are a bit misaligned

martinalong:
You may need to tweak the z index of the <Alert> component in the global carousel so that it will show up above the jumper blur. B/c right now it is hidden by the blur

martinalong:
image
when something is an image you shouldn't be allowed to add an object preview. Use the validations.isImageType function to check the type

PR #973, opened 2021-10-28
Add tags to the filter sidebar

Aminejvm:
This PR updates the filtering sidebar by adding tags and removing both file types and browsers filters. It also adds a responsive popup to access filters alongside the current sidebar

┆Issue is synchronized with this [Notion page](https://www.notion.so/Add-tags-to-the-filter-sidebar-ce24e2908401402489834fc09711fcda) by [Unito](https://www.unito.io)


martinalong:
image
Looks like there's a gap betwen the two headers

martinalong:
image
This arrow should be flipped. It shoudl point up when the menu is open and point down when the menu is closed

martinalong:
image
Let's make sure we truncate long names, and that the icon doesn't get pushed off the side for long names

martinalong:
Can we make the minimum width of the pinned sidebar wider? Like 30-50% wider. It's a bit narrow and squished right now

image

martinalong:
image
There should be some extra padding down at the bottom of the sidebar, so that you can always scroll up enough that the collections aren't covered up by the intercom button

martinalong:
image
If I'm logged in and I'm on a page like this, the sidebar should still be there. Same goes for the new profile page design. You'll always see the top bar and sidebar on each page (only exception being maybe the API page). 

Using the search bar while in a collection will search within that collection (this search option is already supported. Just pass tagIds: ["tag-id-here"] into the search props. Using the search bar while on a profile will search that user's public collections and files (this one is a search option I haven't added yet. I'll add that in a bit)

martinalong:
Looks good! Could you rebase to resolve the conflicts in ApplicationHeader.js?

martinalong:
One thing I think we should do by default is save whether the sidebar is collapsed or not as a setting in the database (and update it as they collapse/uncollapse it). I think people will prefer it one way or the other, and want it to stay that way. B/c I noticed that when I open slate again, it's always collapsed and I have to uncollapse it.

And along with that, let's make sure that for a new user, it defaults to uncollapsed, so that the user can see that this sidebar exists and can see what its purpose is

Collapsed:
image

Uncollapsed:
image


Note: i don't mean update it when people toggle between these two states. Both of these count as the "collapsed" state:
image
image

PR #980, opened 2021-11-09
Fixes: carousel changes

Aminejvm:
fixes
changes requested are in [slate notion](https://slatesystem.notion.site/Carousel-view-2-0-e0c5b3291b03405b8dddc34e8c1245db) and [Carousel PR](https://github.com/filecoin-project/slate/pull/972)

* [x] use medium weight for the jumper header
* [x] change Jumper overlay background color to`bgBlurDark`
* [x] use `borderGrayLight4` for all borders within the jumpers
* [x] create a component to fix the 100vh issue on mobile 
* [x] remove title and description from Link Carousel View

┆Issue is synchronized with this [Notion page](https://www.notion.so/Fixes-carousel-changes-fdc60362cea44920bf5eca5e418319a3) by [Unito](https://www.unito.io)


martinalong:
image
I seem to be getting this error

martinalong:
I went and fixed this for all SVGs by just writing it into the SVG definition, so no need to do this anymore, but just a good think to know:

One thing I'd recommend doing for all SVGs is setting "display: block". This is because without that, there is an extra space below the SVG. For example here, the SVG is 16 x 16, but the immediate div surrounding it is 32 x 35 because of this extra space.

image

image

This is the reason the SVGs in the top bar are slightly misaligned with the view site button

image

Explanation here: https://stackoverflow.com/questions/24626908/how-to-get-rid-of-extra-space-below-svg-in-div-element#:~:text=The%20explanation%20is%20that%20inline,%2C%20'g'%20etc).

martinalong:
Can we make it so the share button is in the carousel header for file owners as well? It's definitely a function that file owners would want (to be able to share their files). And then move the download button from the info jumper to the share jumper? It makes more sense there 

image

image

martinalong:
People who are not the owner of a file shouldn't see this option to select a preview image. We can either remove the section altogether or make the image just display the preview image (or show "no preview image" if there is none)

image

martinalong:
Let's move the download button from the info jumper to the share jumper? It makes more sense there 

image

image

PR #981, opened 2021-11-09
Settings modal jumper

jasonleyser:
Component for new settings jumper, including the modal structure, sidebar and the first profile tab
- settings jumper, open/closing
- sidebar with avatar, username, email and tab buttons
- profile tab with avatar upload, display name and bio change
- replaced the settings link in the drop down menu to open the modal

Screen Shot 2021-11-09 at 1 32 03 PM

### DIFF
--- a/components/core/Settings/Jumper.js
+++ b/components/core/Settings/Jumper.js
@@ -1,0 +1,394 @@
+import * as React from "react";
+import * as Jumper from "~/components/core/Jumper";
+import * as System from "~/components/system";
+import * as Logging from "~/common/logging";
+import * as Strings from "~/common/strings";
+import * as Styles from "~/common/styles";
+import * as Constants from "~/common/constants";
+import * as SVG from "~/common/svg";
+import * as Validations from "~/common/validations";
+import * as Events from "~/common/custom-events";
+import * as Actions from "~/common/actions";
+import * as UserBehaviors from "~/common/user-behaviors";
+
+import { css } from "@emotion/react";
+import { useSettingsContext } from "~/components/core/Settings/Provider";
+import { AnimatePresence, motion } from "framer-motion";
+
+import ProfilePhoto from "~/components/core/ProfilePhoto";
+
+import { Input } from "~/components/system/components/Input";
+import { Textarea } from "~/components/system/components/Textarea";
+import { ButtonPrimary, ButtonSecondary } from "~/components/system/components/Buttons";
+import * as Type from "~/components/system/components/Typography";
+
+const STYLES_FLEX_ROW = `
+  display: flex;
+  flex-direction: row;
+  position: static;
+`;
+
+const STYLES_FLEX_COLUMN = `
+  display: flex;
+  flex-direction: column;
+  position: static;
+  align-items: flex-start;
+`;
+
+const STYLES_JUMPER_HEADER = css`
+  ${Styles.HORIZONTAL_CONTAINER_CENTERED};
+  padding: 17px 20px 15px;
+`;
+
+const STYLES_JUMPER_OVERLAY = (theme) => css`
+  position: fixed;
+  width: 100%;
+  height: 100%;
+  z-index: ${theme.zindex.jumper};
+  @supports ((-webkit-backdrop-filter: blur(75px)) or (backdrop-filter: blur(75px))) {
+    -webkit-backdrop-filter: blur(75px);
+    backdrop-filter: blur(75px);
+    background-color: ${theme.semantic.bgBlurLightTRN};
+  }
+`;
+
+const STYLES_JUMPER_DISMISS_BUTTON = (theme) => css`
+  ${Styles.BUTTON_RESET};
+  color: ${theme.semantic.textGray};
+`;
+
+const STYLES_JUMPER_MAIN = css`
+  ${STYLES_FLEX_ROW}
+  justify-content: space-between;
+  align-items: flex-start;
+  width: 100%;
+  height: 348px;
+  order: 1;
+`;
+
+const STYLES_JUMPER_SIDEBAR = css`
+  ${STYLES_FLEX_ROW}
+  align-items: flex-start;
+  width: 190px;
+  height: 344px;
+  padding: 20px 0px;
+  border-right: 1px solid ${Constants.system.grayLight5};
+`;
+
+const STYLES_JUMPER_CONTENT = css`
+  ${STYLES_FLEX_COLUMN}
+  align-items: flex-start;
+  width: 100%;
+  height: 340px;
+  overflow: scroll;
+  margin: 0px 20px;
+  padding-bottom: 140px;
+  padding-top: 20px;
+
+  ::-webkit-scrollbar {
+    width: 0;
+    background: transparent; 
+  }
+`;
+
+const STYLES_ACCOUNT = css`
+  ${STYLES_FLEX_COLUMN}
+  align-items: center;
+  width: 100%;
+`;
+
+const STYLES_ACCOUNT_USER = css`
+  ${STYLES_FLEX_COLUMN}
+  align-items: center;
+  margin: 10px 0px;
+`;
+
+const STYLES_ITEM_CONTAINER = css`
+  ${STYLES_FLEX_COLUMN}
+  width: 100%;
+  height: 76px;
+  margin: 20px 0px;
+`;
+
+const STYLES_ITEM_HEADER = css`
+  font-size: 12px;
+  margin-bottom: 8px;
+  color: ${Constants.system.gray};
+`;
+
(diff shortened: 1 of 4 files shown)


martinalong:
image
I think whatever wrapper you put around "settings" erased the styling of the word

martinalong:
image
When the settings jumper opens, it should close the top left popover menu

martinalong:
image
Can you make sure the vertical spacing here matches the design? The spacing between each section is different here when they should all be the same

martinalong:
image
check the border radius in the design for the images. this border radius is too small

martinalong:
image
Border radius on all the buttons and inputs is also different from the designs

martinalong:
Let's make sure that user.body has a length limit of 2000, name has a length limit of 255

martinalong:
The profile image upload doesn't seem to be working right now btw

martinalong:
image
Padding here looks too wide or something. can you double check it against the designs?

martinalong:
Also, we are not going to have the theme tab for now b/c implementing darkmode is pretty involved. So you can just take that out

martinalong:
image
Make sure to truncate really long emails that will overflow onto another line with an ellipsis e.g. reallylongemail@gmail....

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,394 @@
...
+import { AnimatePresence, motion } from "framer-motion";
+
+import ProfilePhoto from "~/components/core/ProfilePhoto";
+
+import { Input } from "~/components/system/components/Input";
+import { Textarea } from "~/components/system/components/Textarea";
+import { ButtonPrimary, ButtonSecondary } from "~/components/system/components/Buttons";
+import * as Type from "~/components/system/components/Typography";
+
+const STYLES_FLEX_ROW = `
```
martinalong, line 25: We have a style in common/styles.js for this called HORIZONTAL_CONTAINER. There's other commonly used styles there. Go check it out and see. 

To use it, import it like:
`import * as Styles from "~/common/styles";`
 and use it like this:
image

image

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+        </div>
+
+        <div css={STYLES_MENU_CONTAINER}>
+          {props.tabs.map((tab, i) => {     
+             return(                 
+               <div css={STYLES_MENU_ITEM} key={i}>
+                 <div css={STYLES_MENU_ITEM_ICON}>
+                  {tab.icon}
+                </div>
+                {tab.title}
```
tarafanlin, line 207: make sure every text is wrapped by a style.
I think this one should be: `<SYSTEM.P2>{tab.title}</SYSTEM.P2>`

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+  return (
+    <div css={STYLES_JUMPER_SIDEBAR}>
+      <div css={STYLES_ACCOUNT}>
+        <ProfilePhoto
+          user={props.viewer}
+          style={{ marginBottom: "4px", borderRadius: '10px' }}
+          size={48}
+        />
+        <div css={STYLES_ACCOUNT_USER}>
+          <p style={{ fontWeight: "500", fontSize: "14px" }}>{props.viewer.name}</p>
```
tarafanlin, line 194: this should be `<SYSTEM.H5 style={{ marginBottom: "2px"}}>{props.viewer.name}</SYSTEM.H5>`

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+          <p style={{ fontWeight: "normal", fontSize: "12px" }}>
```
tarafanlin, line 195: this would be `<SYSTEM.P3>{truncateString(props.viewer.email, 20)}</SYTEM.P3>`

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+`;
+
+const STYLES_MENU_ITEM = css`
+  display: flex;
+  font-weight: 400;
+  font-size: 14px;
+  width: 100%;
+  height: 32px;
+  border-radius: 12px;
+  padding: 8px;
```
tarafanlin, line 152: I think the padding and alignment here is off, refer to design file for this.

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+  }
+}
+
+const Sidebar = (props) => {
+  return (
+    <div css={STYLES_JUMPER_SIDEBAR}>
+      <div css={STYLES_ACCOUNT}>
+        <ProfilePhoto
+          user={props.viewer}
+          style={{ marginBottom: "4px", borderRadius: '10px' }}
```
tarafanlin, line 190: borderRadius would be 12px, refer to design file for this.

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+            value={state.body}
+            placeholder="A bit about yourself..."
+            maxLength="2000"
+            onChange={_handleChange}
+          />
+        </div>
+
+        <div css={STYLES_EDIT_INFO_FOOTER}>
+          <ButtonSecondary 
+            style={{ minHeight: "24px", marginLeft: "auto", borderRadius: '8px' }}
```
tarafanlin, line 336: you would need to add paddings to all secondary buttons in smaller size.
refer to design for the numbers.

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+  display: flex;
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  width: 448px;
+  height: 50px;
+  padding: 12px 20px;
+  background-color: ${Constants.system.white};
+  border-radius: 0px 0px 16px 0px;
+  border-top: 1px solid ${Constants.system.grayLight5};
```
tarafanlin, line 133: in 99% of the case, use semantic color rather than system color. 
This would make things easier if we ever go dark mode in the future.
I think the one should be `${Constants.semantic.borderGrayLight}`

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+
+const STYLES_EDIT_INFO_FOOTER = css`
+  display: flex;
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  width: 448px;
+  height: 50px;
+  padding: 12px 20px;
+  background-color: ${Constants.system.white};
```
tarafanlin, line 131: `${Constants.semantic.bgWhite}`

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+  order: 1;
+`;
+
+const STYLES_JUMPER_SIDEBAR = css`
+  ${Styles.HORIZONTAL_CONTAINER};
+  align-items: flex-start;
+  width: 190px;
+  height: 100%;
+  padding: 20px 0px;
+  border-right: 1px solid ${Constants.system.grayLight5};
```
tarafanlin, line 69: `${Constants.semantic.borderGrayLight}`

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+`;
+
+const STYLES_ITEM_CONTAINER = css`
+  ${STYLES_FLEX_COLUMN}
+  width: 100%;
+  height: 76px;
+  margin-top: 24px;
+`;
+
+const STYLES_ITEM_HEADER = css`
```
tarafanlin, line 106: I think you could use `<SYSTEM.H6>` rather than a separate style.

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+  return (
+    <>
+      <div css={STYLES_JUMPER_CONTENT}>
+        <div css={STYLES_AVATAR_BOX}>
+          <p css={STYLES_ITEM_HEADER}>Avatar</p>
+          <div css={STYLES_AVATAR_UPLOAD}>
+            <ProfilePhoto
+              user={state}
+              size={48}
+              style={{ borderRadius: '10px' }}
```
tarafanlin, line 290: border radius is 12px in design

components/core/Settings/Jumper.js
```diff
@@ -0,0 +1,412 @@
...
+`;
+
+const STYLES_EDIT_INFO_FOOTER = css`
+  display: flex;
+  position: absolute;
+  bottom: 0;
+  right: 0;
+  width: 448px;
+  height: 50px;
+  padding: 12px 20px;
```
tarafanlin, line 130: `padding: 8px 20px;`

tarafanlin (changes requested):

PR #1037, opened 2021-12-13
Feat: Onboarding

### LINKED ISSUE #991: Onboarding mobile & revision
┆Issue is synchronized with this [Notion page](https://www.notion.so/Onboarding-mobile-revision-28f4a6efcfbe452a88eae3df566fee37) by [Unito](https://www.unito.io)
┆Figma: https://www.figma.com/file/pi6zKFgcVzeeJiponhiArz/?node-id=586%3A8296

Aminejvm:
This PR resolves #991
For the extension button to work, you'll need to add the following variables to your `.env` file (updated in the [notion doc](https://www.notion.so/slatesystem/Env-variables-42aa05a8f1884d0c89cca9838c48111d)  )

```
NEXT_PUBLIC_EXTENSION_CHROME=https://chrome.google.com/webstore/detail/slate/gloembacbehhbfbkcfjmloikeeaebnoc
NEXT_PUBLIC_EXTENSION_FIREFOX=https://addons.mozilla.org/en-US/firefox/addon/slate-for-firefox
NEXT_PUBLIC_EXTENSION_SAFARI=''
```

### New Features
Codewise the onboarding feature is divided into three main parts, survey (`scenes/SceneSurvey.js`), upload (`components/core/Onboarding/Upload`), and tags (`components/core/Onboarding/Tags.`) onboarding. 
These onboardings will be triggered one after another, its state is managed via the onboarding property in the `viewer` object. The onboarding response will be stored as follows

```jsonc
{
  // What do you currently use for saving things on the web?
  "prevToolsDropbox": false,
  "prevToolsArena": false,
  "prevToolsPinterest": false,
  "prevToolsGoogleDrive": true,
  "prevToolsBrowserBookmarks": false,
  "prevToolsOther": "Raindrop",
  
  // What are you interested in using Slate for?
  "useCasesPersonalStorage": false,
  "useCasesPublicFileSharing": true,
  "useCasesArchiving": false,
  "useCasesBookmarking": false,
  "useCasesMoodboarding": false,
  "useCasesOther": "Research",
  
  // How did you find out about Slate?
  "referralFriend": true,
  "referralTwitter": false,
  "referralIpfsFilecoinCommunity": false,
  "referralOther": "Email",
  
  // onboarding state 
  "surveyCompleted": true,
  "uploadCompleted": true,
  "tagsCompleted": true
}

```

https://user-images.githubusercontent.com/39884652/145868622-bca458af-abff-4345-b407-033500e0e716.mov

### Note
* Currently there are no validations in the survey forms

### Additional changes
* Updated auth popup to the new design
* Integrated auth with the signup section in the landing page
* Fixed `EmptyState`'s layout
* readded the missing auth code
* Moved Jumper and MobileJumper to `/system/fragments`
* Added tags action to ObjectPreview

┆Issue is synchronized with this [Notion page](https://www.notion.so/Feat-Onboarding-23e890a02cd04b73a39955f0ba026808) by [Unito](https://www.unito.io)


martinalong:
Looks phenomenal! I love look of the new empty state and onboarding tour

martinalong:
For the extension url env variables, I actually think it makes more sense to have these as constants in the constants file since they are always the same across dev and production, and we want to be able to easily update them for everyone without having to ask everyone to update their env variables.

The reason we have things like NEXT_PUBLIC_URI_SHOVEL and NEXT_PUBLIC_URI_SLATE in the env file is so that we can give different values for these between dev and production, but this isn't necessary with the extension URL, so it's easier to have it as a constant value that we can just update centrally by pushing new code.

```
NEXT_PUBLIC_EXTENSION_CHROME=https://chrome.google.com/webstore/detail/slate/gloembacbehhbfbkcfjmloikeeaebnoc
NEXT_PUBLIC_EXTENSION_FIREFOX=https://addons.mozilla.org/en-US/firefox/addon/slate-for-firefox
NEXT_PUBLIC_EXTENSION_SAFARI=''
```

martinalong:
On the topic of deciding how to store survey responses in the database, I think it makes more sense as a separate table, where each answer is a separate column. This is for a few reasons: 
- easier and faster to write queries for things at the top level rather than for things inside a jsonb. So it's easier when we are writing our metabase queries
- we want to minimize how much data is in the user table because all of this is selected and returned frequently as users access their account and as others view their stuff. And so we want this to be as small as possible, aka keep any info that doesn't need to be retrieved alongside it in a separate table. survey responses is something that doesn't need to be retrieved alongside user data while using slate
- we also want to be able to version survey, in case we decide to have different questions or answer options later. We don't need to add that column right now, but if we ever add a second version, we can easily just add a version column to the surveys table. If we had this as a jsonb inside the users table, we'd have to use javascript to go through each user and update the jsonb to add a version flag. It's almost always easier to update/manipulate things that are database columns rather than contained inside a jsonb column. For the same reason, having it as columns makes it easier to "remove" an answer by just deleting the column, rather than using javascript to go through and delete it from each jsonb 

Generally you don't want to use a jsonb unless the smallest unit of the thing you're storing IS an object. Only thing we really use it for now is a cover image for a slate or a file, which requires you to store an object item. Because the single item itself IS an object, we have to store it as a jsonb. But for anything where we can "flatten" it and store it as a column, we do.

Proposed structure:
- separate surveys table
- has an ownerId field linked to the id field of users
- individual columns for each of the answers, as booleans. E.g. a column for prevToolsDropbox, prevToolsArena
- to track whether or not a user has answered the survey, you can have a boolean column in `users` called something like "hasCompletedSurvey". This way you can check with the already fetched user data whether to show them the survey, without needing to make a second call to check whether an entry exists for them in surveys

Just make sure you include the script to add this table and its columns in scripts/adjust.js and scripts/seed-database.js so I can modify the other dbs, and so that we keep the seed database script up to date

martinalong:
Upon login, it seems to redirect to /_/404 instead of /_/data

martinalong:
Can you revert the changes you made to the signed out state for application header? It seems to have removed application header when signed out:
image

I think what's already in main should be functioning well for the signed out case already:
image

Aminejvm:
> On the topic of deciding how to store survey responses in the database, I think it makes more sense as a separate table, where each answer is a separate column. This is for a few reasons:
> 
> * easier and faster to write queries for things at the top level rather than for things inside a jsonb. So it's easier when we are writing our metabase queries
> * we want to minimize how much data is in the user table because all of this is selected and returned frequently as users access their account and as others view their stuff. And so we want this to be as small as possible, aka keep any info that doesn't need to be retrieved alongside it in a separate table. survey responses is something that doesn't need to be retrieved alongside user data while using slate
> * we also want to be able to version survey, in case we decide to have different questions or answer options later. We don't need to add that column right now, but if we ever add a second version, we can easily just add a version column to the surveys table. If we had this as a jsonb inside the users table, we'd have to use javascript to go through each user and update the jsonb to add a version flag. It's almost always easier to update/manipulate things that are database columns rather than contained inside a jsonb column. For the same reason, having it as columns makes it easier to "remove" an answer by just deleting the column, rather than using javascript to go through and delete it from each jsonb
> 
> Generally you don't want to use a jsonb unless the smallest unit of the thing you're storing IS an object. Only thing we really use it for now is a cover image for a slate or a file, which requires you to store an object item. Because the single item itself IS an object, we have to store it as a jsonb. But for anything where we can "flatten" it and store it as a column, we do.
> 
> Proposed structure:
> 
> * separate surveys table
> * has an ownerId field linked to the id field of users
> * individual columns for each of the answers, as booleans. E.g. a column for prevToolsDropbox, prevToolsArena
> * to track whether or not a user has answered the survey, you can have a boolean column in `users` called something like "hasCompletedSurvey". This way you can check with the already fetched user data whether to show them the survey, without needing to make a second call to check whether an entry exists for them in surveys
> 
> Just make sure you include the script to add this table and its columns in scripts/adjust.js and scripts/seed-database.js so I can modify the other dbs, and so that we keep the seed database script up to date

Thank you for the proposed structure, it fixed an issue I had previously with the table structure. To verify whether the user submitted the answers or not, I needed an additional request to the surveys table, which will slow the startup performance. But as you mentioned we can avoid it by storing the survey completion state in the users' table and the answers in another table. I'm worried that they may get out of sync if one of the requests fails, but I don't think it will have an impact on the user's experience. 
I also stored the onboarding completion state as columns `hasCompletedUploadOnboarding`, `hasCompletedSlatesOnboarding` in the users table. But I kept the onboarding column since we use it for the privacy alert

scripts/adjust.js
```diff
@@ -42,7 +42,25 @@ const addSlateCoverImage = db.schema.table("slates", function (table) {
   table.jsonb("coverImage").nullable();
 });
 
-Promise.all([addSlateCoverImage]);
+const addSettingsColumnToUsersTable = db.schema
```
martinalong, line 45: What was the reason for adding a settings column? I removed all of these things earlier or moved them into another column already

scripts/adjust.js
```diff
@@ -42,7 +42,25 @@ const addSlateCoverImage = db.schema.table("slates", function (table) {
...
+  .table("users", (table) => {
+    table.jsonb("settings").defaultTo("{}");
+  })
+  .then(() =>
+    db("users").update(
+      "settings",
+      db.raw("jsonb_build_object('filecoin', users.data -> 'settings' )")
+    )
+  );
+
```
martinalong, line 55: Also don't forget to add the createSurveysTable function to the adjust script so that I can run it for other files: 
```
const dropSurveysTable = db.schema.dropTableIfExists("surveys");

const createSurveysTable = createTableIfNotExists("surveys", function (table) {
  table.uuid("id").primary().unique().notNullable().defaultTo(db.raw("uuid_generate_v4()"));
  table.uuid("ownerId").references("id").inTable("users");

  // What do you currently use for saving things on the web?
  table.boolean("prevToolsDropbox").defaultTo(false);
  table.boolean("prevToolsArena").defaultTo(false);
  table.boolean("prevToolsPinterest").defaultTo(false);
  table.boolean("prevToolsGoogleDrive").defaultTo(false);
  table.boolean("prevToolsBrowserBookmarks").defaultTo(false);
  table.string("prevToolsOther").defaultTo(null);

  // What are you interested in using Slate for?
  table.boolean("useCasesPersonalStorage").defaultTo(false);
  table.boolean("useCasesPublicFileSharing").defaultTo(false);
  table.boolean("useCasesArchiving").defaultTo(false);
  table.boolean("useCasesBookmarking").defaultTo(false);
  table.boolean("useCasesMoodboarding").defaultTo(false);
  table.string("useCasesOther").defaultTo(null);

  // How did you find out about Slate?
  table.boolean("referralFriend").defaultTo(false);
  table.boolean("referralTwitter").defaultTo(false);
  table.boolean("referralIpfsFilecoinCommunity").defaultTo(false);
  table.string("referralOther").defaultTo(null);
});
```

scripts/adjust.js
Aminejvm: I added it back so I can store the filter sidebar state, I feel like storing it as a column will clutter the users' table. There is another option, we can create a settings table. I can also add to it the state of the sidebar sections if each section is collapsed or not

martinalong:
Just double checked with haris and we wanted to update some of the copy and options for the survey:

Q1: How do you currently save things on the web?
Browser bookmarks
Pinterest
Arena
Notes platform (Notion, Evernote, etc.)
Other

Q2: What are you interested in using Slate for?
Bookmarking important pages
Saving links to read later
Searching your browsed pages
Sharing collections of links
Other

Q3: Can stay the same

Welcome Screen: "Slate is a personal search engine that's designed to help you save and remember things you care about on the web.

martinalong:
image

image

I'm experiencing an issue which keeps the page from loading, but only occurs if you navigate to the data page when already authenticated, and only for a user that has not completed the survey yet.

If you are not logged in and go through the login page first, it works fine.
And once you've completed the survey + onboarding, it works fine. So I suspect it's something to do with the survey/onboarding

PR #1046, opened 2022-01-04
Improvements and accessibility changes

### LINKED ISSUE #1032: Misc remaining tasks for Amine
┆Issue is synchronized with this [Notion page](https://www.notion.so/Misc-remaining-tasks-for-Amine-647f69a40fc54cb5bdab98186a5c3334) by [Unito](https://www.unito.io)

Aminejvm:
Branched form #1037 
This PR resolves #1032
### New Features
* Added FocusRing component

Displays a focus ring if its child receives focus via a keyboard. This component doesn't create a dom node, it uses its child component to detect focus events. Therefore the child component forward its ref
```jsx
const Button = (props) => (
  <FocusRing>
    <button {...props} />
  </FocusRing>
);
```
FocusRing can also detect focus within its children, similar to the `:focus-within` selector. You can toggle it by using the `within` prop

```jsx
const STYLES_FOCUS_RING = css`
  outline: 2px solid red;
`;

const Input = () => (
  <FocusRing within css={STYLES_FOCUS_RING}>
    <input  />
    <button>
      <Icon />
    </button>
  </FocusRing>
);
```

* RovingTabIndex component

Navigate a list's children using the arrow keys. Navigation's direction is managed by the `axis` prop, `horizontal` using the left and right arrows, or `vertical` with the up and down arrows. `RovingTabIndex.Item` should wrap directly the focusable element, and its child should forward its ref. 
This component manages the navigation order by keeping track of each focusable element in the list using the `index` prop. Upon hitting the arrow key, we'll update the focusable element's tabindex to 0 and focus it manually

```jsx
const Navigation = () => (
  <RovingTabIndex.Provider axis="vertical">
    <RovingTabIndex.List as="ul">
      {items.map((item, index) => (
        <li key={item.id}>
          <RovingTabIndex.Item index={index}>
            <button>{item.name}</button>
          </RovingTabIndex.Item>
        </li>
      ))}
    </RovingTabIndex.List>
  </RovingTabIndex.Provider>
);
```

Currently works in the tags jumper and in the filter sidebar and popup.

https://user-images.githubusercontent.com/39884652/148104365-dd2b161f-7b4d-4993-8ee6-6ae84bfd32b5.mov

* New Tooltip component

Follow guidelines from [wai-aria/tooltip](https://www.w3.org/TR/wai-aria-practices/#tooltip) . Similar to the jumper component, you need to import all parts to build a tooltip. Tooltip's placement is managed by the `vertical` and `horizontal` props.
Similar to previous components, `Tooltip.Trigger` doesn't create a dom node, so its child should forward its ref
```jsx
const ShortcutTooltip = () => (
  <Tooltip.Root vertical="below" horizontal="right">
    <Tooltip.Trigger aria-describedby="trigger-tooltip">
      <button>
        <Icon />
      </button>
    </Tooltip.Trigger>
    <Tooltip.Content>
      <System.H6 id="trigger-tooltip" as="p" color="textGrayDark">
        Click to save
      </System.H6>
    </Tooltip.Content>
  </Tooltip.Root>
);
```

### Additional changes
* Toggle the filter sidebar via `\` key
* Ability to collapse and expand filters
* Added keyboard shortcuts for carousel's actions
     * `e` for edit info jumper
     * `s` for share jumper
     * `t` for tags jumper
     * `i` for more info jumper
* Show different search scopes in the filter's header
* Accessibility changes to the carousel and jumpers following [wai-aria](https://www.w3.org/TR/wai-aria-practices-1.1/) guides

### DIFF
--- a/common/hooks/a11y.js
+++ b/common/hooks/a11y.js
@@ -1,0 +1,55 @@
+import * as React from "react";
+
+import { first, last } from "lodash";
+import { useEventListener } from "~/common/hooks";
+
+// SOURCE(amine): https://zellwk.com/blog/keyboard-focusable-elements/
+const getFocusableElements = (element = document) =>
+  [
+    ...element.querySelectorAll(
+      'a[href], button, input, textarea, select, details,[tabindex]:not([tabindex="-1"])'
+    ),
+  ]
+    // NOTE(amine): remove disabled buttons and elements with aria-hidden attribute
+    .filter(
+      (el) =>
+        !el.hasAttribute("disabled") &&
+        !el.getAttribute("aria-hidden") &&
+        el.getAttribute("tabindex") !== "-1"
+    );
+
+/* NOTE(amine): used to trap focus inside a component. **/
+export const useTrapFocus = ({ ref }) => {
+  const handleFocus = (e) => {
+    if (!ref.current) return;
+    const elements = getFocusableElements(ref.current);
+    const firstElement = first(elements);
+    const lastElement = last(elements);
+
+    const isTabPressed = e.key === "Tab" || e.keyCode === "9";
+    if (!isTabPressed) return;
+
+    if (e.shiftKey) {
+      if (document.activeElement === firstElement) {
+        lastElement.focus();
+        e.preventDefault();
+      }
+    } else {
+      if (document.activeElement === lastElement) {
+        firstElement.focus();
+        e.preventDefault();
+      }
+    }
+  };
+
+  useEventListener({ type: "keydown", handler: handleFocus }, []);
+};
+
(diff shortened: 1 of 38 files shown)


martinalong:
image
image
Seems like the positioning of the tagging button got messed up for file (non-link) objects

martinalong:
https://www.loom.com/share/4f2ceaae2c7540849ff74da35b598240

For the tagging interface, left/right arrow keys navigates the different tags, but it also triggers the "next object" "previous object". Let's make sure to add a stop propagation on this to stop it from also changing the object when the jumper is open

martinalong:
I can't seem to select the objects to open them with the carousel. Let's make it so tabbing through these will select the individual objects, rather than just the external links?
image

PR #1047, opened 2022-01-04
Improvements and bug fixes

Aminejvm:
Branched from #1046 

### Changes
- [x] display loader spinner when saving a link
- [x] Make the toolbar persist in the carousel
- [x] Correctly reset upload state when an upload finishes
- [x] Fix a bug where setting `new Image().src` to null redirect the app to `/_/null`
- [x] Hide the tags keyboard shortcut when a tag is just created
- [ ] Merge search and header bar into one section
- [ ] Add a box-shadow to popup

### DIFF
--- a/common/hooks/a11y.js
+++ b/common/hooks/a11y.js
@@ -1,0 +1,55 @@
+import * as React from "react";
+
+import { first, last } from "lodash";
+import { useEventListener } from "~/common/hooks";
+
+// SOURCE(amine): https://zellwk.com/blog/keyboard-focusable-elements/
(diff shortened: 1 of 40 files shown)


martinalong:
image
I think you may have branched off an older branch for this one, and not main? Bc it pulled in a lot of older, unrelated commits. It's fine for this one, but just a heads up for next time

martinalong:
Looks good! Can we also add to this the clearing of the input bar after you click create tag? E.g here after i clicked "create tag", the input bar should be cleared and it should go back to the default all tags view in the second image:
image

image

